Issue #8, opened 2016-06-02
Provide a way to edit existing facts.


elbenfreund:
``hamster-time-tracker`` provides a dedicated ``edit`` dialog to change a ``Fact``s attributes.
This should be reimplemented.

Questions:
Is it desirable to use the ``raw_fact`` entry to modify the fact (but for the description?)
It strikes me as rather cumbersome. Wouldn't it be preferable to have dedicated widgets for each 
fact attribute? What do you think @projecthamster/devs ?

elbenfreund:
Closed in PR #23

elbenfreund:
A bug in ``hamster-lib`` does prevent updating a fact with unchanged start/end values.

PR #36, opened 2016-06-22
#8 Provide an 'edit existing fact' dialog

elbenfreund:
#Dialog opens when a row in the overview is ``activated``.

This PR also makes sure that the ``HamsterGTK`` application is passed
along to all major players in a consistent and transparent way instead
of hoping to be able to access it as some parent window attribute.

A thing to notice is: The string used to populate the ``raw fact`` aka
``activity`` entry widget contains a proper raw fact now. That is a
legacy ``serialized_name`` prepended with the facts time info.
Another difference is that ``time frame`` is given in the proper format
where ``start`` and ``end`` are seperated by an ``-`` not by an `` `` as
in the legacy version. Whilst this breaks the previous string, it should
be alright and gets us one step closer to higher code consistency.

A new ``SignalHandler`` class is added that can be used to register
generic custom signals. This may not be the idiomatic way to do this but
then agian, is rather confusing finding any consistent information on
this matter to begin with. Instead of creating a new 'controler` class
inheriting from ``HamsterControler`` *and* ``GObject`` we opted for
``composition`` instead.  An instance of this class is added as
``HamsterGTa.controler.signal_handler`` and provides an easy way to
emit custom signals.

With regards to ``EditFactDialog.apply``:
Instead of parsing the ``raw_fact`` and tranfering the attributes to
``self._fact`` we cheat a bit and stick with the new instance and just
attach the PK of the fact to be updated before we pass it off to
``hamsterlib``.

This PR does not include any visual representation day representation like the original hamster did (``widgets.DayLine()``). This will be added in a separate issue.

Closes: #8

### DIFF
--- a/hamster_gtk/__init__.py
+++ b/hamster_gtk/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-"""hamster-gtk is a Gtk based timetracking GUI build om top of ``hamsterlib``."""
+"""hamster-gtk is a Gtk based timetracking GUI build om top of ``hamster-lib``."""
 
 __author__ = 'Eric Goller'
 __email__ = 'Elbenfreund@DenkenInEchtzeit.net'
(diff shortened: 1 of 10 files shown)


elbenfreund:
Whilst the PR works, the underlying ``hamster-lib`` update method is broken right now. The GUI deal with this properly by rendering an error dialog. Nevertheless, in order to see it actually working we will wait with the merge until ``hamster-lib`` is fixed.

Issue #42, opened 2016-07-07
i18n and l10n


elbenfreund:
One basic core functionality has been implemented, we need to address i18n and l10n.
It is advisable not to postpone this for two long in order to minimize duplicate effort.

It appears we can not simply use gettext for direct GTK related translations. This needs to be investigated!

Task related to i18n:
* [ ] Provide i18n for all GUI labels.
* [ ] Decide on a solution for i18n within the logging.
* [ ] Provide i18n setup for documentation.

Tasks related to i10n:
* [ ] Localize datetime representation #41

jtojnar:
We could also use [l20n](https://github.com/l20n/python-l20n), which supports much more language features than gettext. Unfortunately, the tool support is currently [not that great](https://github.com/translate/translate/milestone/24).

PR #75, opened 2016-09-13
Move UI to resource files

jtojnar:
After resolving #4, I have decided to port the UI to Gtk.Builder/Glade. Doubtlessly, this separation will lead to cleaner and more maintainable code, basically [i18n](https://github.com/projecthamster/hamster-gtk/issues/42) for free!

Currently only the Preferences dialog is ported, which I also polished slightly; other parts of the application can follow later.

In the future, other than porting more modules, we should look into handling the files as [glib resources](https://developer.gnome.org/gio/2.32/gio-GResource.html), if it is even possible without the autotools mess.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,6 +9,7 @@ include .editorconfig
 
 recursive-include tests *
 recursive-include hamster_gtk *.py
+recursive-include hamster_gtk/resources/css *.css
 recursive-include requirements *.pip
 recursive-include docs *.rst conf.py Makefile make.bat
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -10,6 +10,7 @@ include .editorconfig
 recursive-include tests *
 recursive-include hamster_gtk *.py
 recursive-include hamster_gtk/resources/css *.css
+recursive-include hamster_gtk/resources/ui *.ui
 recursive-include requirements *.pip
 recursive-include docs *.rst conf.py Makefile make.bat
 
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -38,6 +38,7 @@ import hamster_lib
 from backports.configparser import SafeConfigParser
 from gi.repository import Gdk, GObject, Gtk
 from hamster_lib.helpers import config_helpers
+from hamster_gtk.helpers import get_resource_path
 from six import text_type
 
 # [FIXME]
@@ -117,7 +118,7 @@ class MainWindow(Gtk.ApplicationWindow):
 
         # Setup css
         style_provider = Gtk.CssProvider()
-        style_provider.load_from_data(self._get_css().encode())
+        style_provider.load_from_path(get_resource_path('css/hamster-gtk.css'))
         Gtk.StyleContext.add_provider_for_screen(
             Gdk.Screen.get_default(),
             style_provider,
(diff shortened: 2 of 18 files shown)


elbenfreund:
I am about to call it a night, but wanted to give some initial feedback right away.
First of all, thanks for the effort and initiative.
I deliberatly opted against a glade based aproach,something I will try to elaborate on tomorrow. So this is just a heads up.
In general, please feel invited and encouraged to join the conversation about general design / direction questions on the [mailing list](http://projecthamster.org/pages/get-in-touch.html)

elbenfreund:
Hello jtojnar,

my sincere apologies for this delayed feedback. Contrary to what you may have deducted from my previous comment and long delay I am actually very excited about your work and its high degree of quality.

Before I try to elaborate in some more details about my thoughts on glade in general let me provide some simple PR specific comments first.
* Each PR should deal with one issue only. As a consequence this one PR should be split into two:
  * One dealing with #4 
  * One about refactoring the dialog into something glade based (with a corresponding issue).
* The new helper function ``get_resource_path`` should be tested, despite being rather trivial.

#### With regards to your refactoring of the Preference dialog class
The two main changes beyond CSS are not merely moving to glade but just as important adding dedicated custom widget classes where the previous preferences dialog used simple private methods that returned customized widget instances.
* From this perspective it's more of an implementation detail that those new custom widget classes don't define the general layout in code but use glade/builder.
* In general I am a huge fan of custom classes and they are used liberally in the codebase. Where  they really shine is in encapsulating custom functionality within the widget. As the preferences dialog really is just about arranging default widgets with relatively straight forward two liners having trivial classes seemed a bit overkill. Instead "encapsulation" for clean code is achieved by using lots of separate helper methods.
* My main point boils down to: glade is only viable for trivial GUIs. But those are just as easily written as plain python files. This is not where the savings live. The real work in done in more complex widgets that glade can not deal with to begin with.
* I am not sure that the approach taken so far is that unmaintainable. I absolutely agree it requires a lot of discipline not to build pages of layout constructing code within one big scope but this has been avoided by using small, well defined and specialized private methods.

So, to sum this up:
Right now, I am not really convinced that using glade saves us too much work and provides another layer. Besides that however I can see (and share) your approach of having specialized custom widget classes. This is one major tool for keeping our GUI code clean.
In the particular case of the Preference dialog I personally felt it was not yet needed and simple encapsulation by dedicated function was modular enough. But I can easily understand if one end up on the other end of that argument.

#### Bottom line
I would be very glad to work with you on hamster/hamster-gtk. I feel migrating to Glade is not a top priority at this point, especially as most of the rough GUI code is already there, so there is not much work that could be saved. In the mid/long term, ``hamster-gtk`` is looking for a dedicated maintainer who, as the person having to maintain it, would have a major word in deciding whether he/she prefers glade or plain python code.
I hope you are not too put off by me not wholeheartedly merging your contribution and look forward to your feedback. Whilst I so far failed to reply to you on IRC in time, I still think this may be a better place for detailed discussions and will try to catch you there the next couple of days.

#### Appendix: Glade in general
* GTK documentation is shit to begin with. Having another layer on top of the actual C reference makes things worse. At the same time it encourages use of "options" and "settings" without proper understanding of the underling class and its properties.

* It adds another layer of code<->UI declaration interaction. It is no longer possible to just browse the code in order to understand whats going on. Whilst the XML is indeed very readable it makes things less readable none the less as definitions, options and settings are distributed over various places.

* Providing context dependent documentation / comments is much harder. Of cause we could edit the raw XML but this can not be expected to be read by the average passing developer.

jtojnar:
Thank you for the comprehensive reply. I know the multi-feature PR is iffy. I submitted the both changes in the same PR because I was not sure how to handle the Glade commits when they depend on the `get_resource_path` from CSS commit.

Now that the CSS commit was merged I will either rebase this PR or submit a new one.

PR #83, opened 2016-09-24
Refactor preferences into separate files

jtojnar:
PreferencesDialog was extracted from main file and the preferences widgets were extracted from the preferences dialog. Additionally, mnemonics were added to preferences.


tests/preferences/conftest.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+@pytest.fixture
+def time_entry(request):
+    """Return a TimeEntry instance."""
+    return widgets.TimeEntry()
+
+
+# Data
+
```
elbenfreund: The global config contains a ``initial_config_parametrized`` fixture that is used for testing preferences only. This and its associated fixtures should probably be moved over here as well.

tests/preferences/test_preferences_dialog.py
```diff
@@ -0,0 +1,27 @@
...
+class TestPreferencesDialog(object):
+    """Unittests for PreferencesDialog."""
+
+    def test_init(self, dummy_window, app, config):
+        """Make instantiation works as expected."""
+        result = PreferencesDialog(dummy_window, app, config)
+        grid = result.get_content_area().get_children()[0]
+        # This assumes 2 children per config entry (label and widget).
+        assert len(grid.get_children()) / 2 == len(config.keys())
+
```
elbenfreund: There  seems to be no test for ``set_config``.

tests/preferences/test_widgets.py
```diff
@@ -0,0 +1,163 @@
...
+    def test_instance(self):
+        """Make sure the widget is still a Grid."""
+        file_chooser = widgets.EditableFileChooser()
+        assert isinstance(file_chooser, Gtk.Grid)
+
+    def test_mnemonic(self, editable_file_chooser):
+        """Make sure the widget can be accessed using mnemonic."""
+        assert editable_file_chooser.on_mnemonic_activate(editable_file_chooser, False)
+
+    def test_set_get(self, editable_file_chooser, paths):
```
elbenfreund: Please provide two independent tests, each testing *only* ``set_config`` or ``get_config``.

tests/preferences/test_widgets.py
```diff
@@ -0,0 +1,163 @@
...
+        spin = widgets.DurationEntry(**spinner_params)
+        assert spin.get_digits() == 19
+
+    def test_params_constructor_clamped_digits(self, spinner_params):
+        """When passing a step with too many decimal places, the precision is clamped."""
+        spinner_params['step'] = 1e-25
+        spin = widgets.DurationEntry(**spinner_params)
+        assert spin.get_digits() == widgets.duration_entry.MAX_DIGITS
+
+    def test_set_get(self, duration_entry, numbers):
```
elbenfreund: Please provide two independent tests, each testing *only* ``set_config`` or ``get_config``.

tests/preferences/test_widgets.py
```diff
@@ -0,0 +1,163 @@
...
+        entry = widgets.TimeEntry()
+        assert isinstance(entry, Gtk.Entry)
+
+    def test_get_invalid(self, time_entry):
+        """Make sure an error is raised when the entered value is invalid."""
+        time_entry.set_text('moo')
+        with pytest.raises(ValueError):
+            time_entry.get_config_value()
+
+    def test_set_get(self, time_entry, times):
```
elbenfreund: Please provide two independent tests, each testing *only* ``set_config`` or ``get_config``.

hamster_gtk/preferences/widgets/config_combo_box_text.py
```diff
@@ -0,0 +1,48 @@
...
+    # Required else you would need to specify the full module
+    # name in ui file
+    __gtype_name__ = 'ConfigComboBoxText'
+
+    def __init__(self, values=None):
+        """Initialize widget."""
+        super(Gtk.ComboBoxText, self).__init__()
+
+        if values is not None:
+            for id, value in values:
```
elbenfreund: If it is true that ``id`` can be an arbitrary string, maybe a better variable name can be chosen. Whilst ID's of cause do not need to be numeric, its a common first thought I think.

hamster_gtk/preferences/widgets/config_combo_box_text.py
```diff
@@ -0,0 +1,48 @@
...
+                self.append(id, value)
+
+    def get_config_value(self):
+        """Return selected value."""
+        return self.get_active_id()
```
elbenfreund: Should't this be wrapped in ``hamster.helpers._u()`` to make sure a unicode instance is returned regardless of python version?

hamster_gtk/preferences/widgets/editable_file_chooser.py
```diff
@@ -0,0 +1,81 @@
...
+        self.button = Gtk.Button(_("Choose"))
+        self.button.connect('clicked', self.on_choose_clicked)
+
+        self.attach(self.entry, 0, 0, 1, 1)
+        self.attach(self.button, 1, 0, 1, 1)
+        self.connect('mnemonic-activate', self.on_mnemonic_activate)
+
+    def get_config_value(self):
+        """Return value of the entry."""
+        return self.entry.get_text()
```
elbenfreund: Should't this be wrapped in ``hamster.helpers._u()`` to make sure a unicode instance is returned regardless of python version?

hamster_gtk/preferences/widgets/config_combo_box_text.py
jtojnar: Here I was going with [GTK terminology](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/ComboBoxText.html#Gtk.ComboBoxText.append). I can change it if needed

elbenfreund: No, you are right. Once the docstring gets expanded things should be clearer anyway.

hamster_gtk/preferences/preferences_dialog.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        super(PreferencesDialog, self).__init__(*args, **kwargs)
+
+        self._parent = parent
+        self._app = app
+
+        self.set_transient_for(self._parent)
+
+        db_engines = [('sqlite', _('SQLite')), ('postgresql', _('PostgreSQL')),
+            ('mysql', _('MySQL')), ('oracle', _('Oracle')), ('mssql', _('MSSQL'))]
```
jtojnar: I am not really convinced about this, for example Japanese might want to transliterate the names to Katakana, Russians might want to use Cyrillic or Thai whatever their script is called – [มายเอสคิวเอล](https://th.wikipedia.org/wiki/%E0%B8%A1%E0%B8%B2%E0%B8%A2%E0%B9%80%E0%B8%AD%E0%B8%AA%E0%B8%84%E0%B8%B4%E0%B8%A7%E0%B9%80%E0%B8%AD%E0%B8%A5). Yay for different cultures.

hamster_gtk/preferences/widgets/config_combo_box_text.py
```diff
@@ -0,0 +1,48 @@
...
+"""This module provides ComboBoxText widget extended to store preferences."""
+
+from __future__ import absolute_import
+
+from gi.repository import Gtk
+
+from .config_widget import ConfigWidget
+
+
+class ConfigComboBoxText(Gtk.ComboBoxText, ConfigWidget):
```
jtojnar: `HamsterComboBoxText` works? I think GIMP does it like this.

jtojnar:
Okay, that should be everything except for renames. I will do that later for diff clarity.

> To be in line with the other submodules, the dialog should be housed in ``.dialogs``.

> Maybe ``ValueAccessMixin`` or ``CustomValueAccessMixin`` would be a better classname?
If the name is changed, we need to adjust the module name.

> Probably better to reflects its rather generic nature by calling it ``CustomSpinButton`` or just ``SpinButton``?

> An alternative would be ``ComboFileChooser``, what do you think?

tests/preferences/conftest.py
```diff
@@ -0,0 +1,95 @@
...
+@pytest.fixture
+def numbers(request, faker):
+    """Return a list of random numbers."""
+    amount = 8
+    return [faker.random_number(digits=5) for i in range(amount)]
+
+
+@pytest.fixture
+def times(request, faker):
+    """Return a list of random numbers."""
```
jtojnar: Not many other fixtures use `random_` prefix. `random_number` is one, but it looks more like an exception rather than a rule.

hamster_gtk/preferences/preferences_dialog.py
elbenfreund: I can live with that reasoning.

elbenfreund:
For those lines where you addes ``Args`` and ``Returns`` sections to docstring, could you please add type references.
For example please refer to [here](http://www.sphinx-doc.org/en/stable/ext/napoleon.html#module-sphinx.ext.napoleon).

tests/preferences/test_preferences_dialog.py
elbenfreund: This seems to be still true in ``2758ea65c4ea930a97192ad0db6d96aeb5df53bd``.

jtojnar: Yes, the `set_config` is now a private method :-D

elbenfreund: shame on me

tests/preferences/widgets/test_hamster_combo_box_text.py
```diff
@@ -0,0 +1,48 @@
...
+        combo_box = widgets.HamsterComboBoxText(combo_box_items)
+        assert len(combo_box.get_model()) == len(combo_box_items)
+
+    def test_get_config_value(self, hamster_combo_box_text, combo_box_items):
+        """Make sure the widget value is retrieved correctly."""
+        random_id = random.choice(combo_box_items)[0]
+        hamster_combo_box_text.set_active_id(random_id)
+        assert hamster_combo_box_text.get_config_value() == random_id
+
+        another_random_id = random.choice(combo_box_items)[0]
```
elbenfreund: Wouldn't it be cleaner to do:
``selected_ids = [item[0] for item in random.sample(combo_box_items), 2)]``
and  then just iterate the test logic over them only once?

tests/preferences/widgets/test_hamster_combo_box_text.py
```diff
@@ -0,0 +1,48 @@
...
+        hamster_combo_box_text.set_active_id(another_random_id)
+        assert hamster_combo_box_text.get_config_value() == another_random_id
+
+    def test_set_config_value(self, hamster_combo_box_text, combo_box_items):
+        """Make sure the widget value is set correctly."""
+        random_id = random.choice(combo_box_items)[0]
```
elbenfreund: see ``test_get_config_value`` about reducing code duplication.

tests/preferences/widgets/test_hamster_spin_button.py
```diff
@@ -0,0 +1,79 @@
...
+        assert adjustment.get_step_increment() == simple_adjustment.step
+
+    def test_params_constructor_adjustment(self, adjustment):
+        """Make sure the widget can be set up by passing :class:`Gtk.Adjustment` to constructor."""
+        spin = widgets.HamsterSpinButton(adjustment)
+        assert spin.get_adjustment() == adjustment
+
+    def test_params_constructor_overlapping_bounds(self, simple_adjustment):
+        """Passing lower bound greater than upper bound to the constructor should fail."""
+        new_adj = simple_adjustment._replace(min=simple_adjustment.max, max=simple_adjustment.min)
```
elbenfreund: Please construct a new ``SimpleAdjustment`` instance from scratch rather than calling private methods.

tests/preferences/widgets/test_hamster_spin_button.py
```diff
@@ -0,0 +1,79 @@
...
+        with pytest.raises(ValueError):
+            widgets.HamsterSpinButton(new_adj)
+
+    def test_params_constructor_zero_step(self, simple_adjustment):
+        """Passing zero as a step to the constructor should fail."""
+        new_adj = simple_adjustment._replace(step=0)
```
elbenfreund: Please construct a new ``SimpleAdjustment`` instance from scratch rather than calling private methods.

tests/preferences/widgets/test_hamster_spin_button.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Passing a value that is neither a :class:`Gtk.Adjustment`, nor a :class:`SimpleAdjustment`
+        to the constructor should fail.
+        """
+        with pytest.raises(ValueError):
+            widgets.HamsterSpinButton(42)
+
+    def test_params_constructor_decimal_places(self, simple_adjustment):
+        """When given a fractional step, the widget should show values at appropriate precision."""
+        new_adj = simple_adjustment._replace(step=1e-19)
```
elbenfreund: Please construct a new ``SimpleAdjustment`` instance from scratch rather than calling private methods.

tests/preferences/widgets/test_hamster_spin_button.py
```diff
@@ -0,0 +1,79 @@
...
+        spin = widgets.HamsterSpinButton(new_adj)
+        assert spin.get_digits() == 19
+
+    def test_params_constructor_clamped_digits(self, simple_adjustment):
+        """When passing a step with too many decimal places, the precision is clamped."""
+        new_adj = simple_adjustment._replace(step=1e-25)
```
elbenfreund: Please construct a new ``SimpleAdjustment`` instance from scratch rather than calling private methods.

tests/preferences/widgets/test_time_entry.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test_get_invalid(self, time_entry):
+        """Make sure an error is raised when the entered value is invalid."""
+        time_entry.set_text('moo')
+        with pytest.raises(ValueError):
+            time_entry.get_config_value()
+
+    def test_get_config_value(self, time_entry, times):
+        """Make sure the widget value is retrieved correctly."""
+        for time in times:
```
elbenfreund: We should also test for the more liberal case: ``%H:%M``.

hamster_gtk/preferences/preferences_dialog.py
```diff
@@ -0,0 +1,112 @@
...
+            parent (Gtk.Window): Dialog parent.
+            app (HamsterGTK): Main app instance. Needed in order to retrieve
+                and manipulate config values.
+        """
+        super(PreferencesDialog, self).__init__(*args, **kwargs)
+
+        self._parent = parent
+        self._app = app
+
+        self.set_transient_for(self._parent)
```
jtojnar: No, it is already used (up the stream](https://github.com/jtojnar/hamster-gtk/blob/b5c3b886f76a602b12c0e0d2b3e439071f747ca0/hamster_gtk/hamster_gtk.py#L101).

hamster_gtk/preferences/widgets/combo_file_chooser.py
```diff
@@ -0,0 +1,93 @@
...
+    __gtype_name__ = 'ComboFileChooser'
+
+    def __init__(self):
+        """Initialize widget."""
+        super(Gtk.Grid, self).__init__()
+
+        self._entry = Gtk.Entry()
+        self._entry.set_hexpand(True)
+
+        self._button = Gtk.Button(_("Choose"))
```
jtojnar: No, I do not use mnemonics here as it would probably be confusing. It would be the same for all the buttons in Preferences and would cycle between them.

People can access the entry using mnemonics and then <kbd>tab</kbd> to use the button. Maybe the button could be focused when `_on_mnemonic_activate` is called, though accessing the entry would then require <kbd>shift</kbd>-<kbd>tab</kbd>.

hamster_gtk/preferences/widgets/time_entry.py
```diff
@@ -0,0 +1,63 @@
...
+
+class TimeEntry(Gtk.Entry, ConfigWidget):
+    """A widget for entering time."""
+
+    # Required else you would need to specify the full module name in ui file
+    __gtype_name__ = 'TimeEntry'
+
+    def get_config_value(self):
+        """
+        Return selected time.
```
jtojnar: Yes, I totally forgot to do this, even though I thought about it several times.

The problem I have with this is that the error is only thrown when the value being accessed instead of when user types it in. Ideally the submission of the form (preferences dialogue) should be disabled until the value is in correct form. This is better left for another PR.

Another way of solving this is replacing the entry with three `SpinButton`s, though that would require some work in making it not suck UX-wise. Also not in the scope of this PR.

elbenfreund: I agree with the best case solution to be input validation as well as with that this should be a separate issue.
However, the docstring should document the actual behavior as is. And that would be raising an ``ValueError``? exception if not matching one of the two formats.

jtojnar:
Ad the removal of Test classes: I want to mirror the library structure with test structure completely, including the classes (i.e. method `foo` of `hamster_gtk.xxx.yyy.zzz.QuuxClass` should be tested by `tests.xxx.yyy.test_zzz.TestQuuxClass.test_foo`).

Additionally, having classes there will make it easier to see what component failed, as pytest uses `TestClass.test_method` heading for a failed test. Compare `TestTimeEntry.test_get_config_value` to `test_get_config_value`.

tests/preferences/test_preferences_dialog.py
jtojnar: It is actually covered with `get_config`, because `_set_config` is used by the constructor. I added [a simple test](https://github.com/jtojnar/hamster-gtk/commit/99e7b55ed0b299f836361c4b7820d08790624446) but it cannot test anything until I get my hands on `another_config_parametrized`. Which I have no idea how to do.

elbenfreund:
Whilst I do disagree with your second paragraph I will let it go like this for now. During the next couple of iterations there will very likely be some restructuring of this code/test anyway, so I opt for moving forward with this.

Issue #46, opened 2016-07-11
Reevaluate the use of Gtk.STOCK* buttons


elbenfreund:
According to [this](http://lazka.github.io/pgi-docs/Gtk-3.0/constants.html#Gtk.STOCK_OK) Stock buttons are deprecated since 3.10. The new solution however is unclear.

PR #86, opened 2016-09-24
[#46] Replace stock buttons with regular buttons

jtojnar:
Closes: #46

### DIFF
--- a/hamster_gtk/misc/dialogs.py
+++ b/hamster_gtk/misc/dialogs.py
@@ -113,7 +113,7 @@ class DateRangeSelectDialog(Gtk.Dialog):
 
     # Widgets
     def _get_apply_button(self):
-        button = Gtk.Button.new_from_stock(Gtk.STOCK_APPLY)
+        button = Gtk.Button(_('_Apply'))
         return button
 
     def _get_today_widget(self):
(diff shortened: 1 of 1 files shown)


elbenfreund:
My understanding of  #46 is that "the right way" (tm) to deal with former stock buttons is not quite clear. Did you find any documentation/source that elaborates on what GTKs new best practice is?
Maybe its worth asking the gnome list...

jtojnar:
The [documentation](https://developer.gnome.org/gtk3/stable/gtk3-Stock-Items.html#GTK-STOCK-OK:CAPS) says we should just use plain label. ebassi explains it in a StackOverflow [answer](http://stackoverflow.com/a/30986414/160386).

elbenfreund:
I knew about the passage in the documentation and I agree that the link indicates that one should use just a labeled button but to me it was not utterly clear if there are drop in replacements for the old stock buttons to be used instead (that would be label only) or if the conveyed meaning is: "just use plain label buttons".

Here is further feedback from ``#gnome-hackers``:
> (13:25:25) zeenix: "gtk_button_new_from_stock has been deprecated since version 3.10 and should not be used in newly-written code.
> (13:25:25) zeenix: Use gtk_button_new_with_label() instead."

Thanks for taking on this issue. I will start the review right away.

jtojnar:
I like using a proper constructor better. `new_*` functions are used in C because it lacks proper classes.

PR #89, opened 2016-09-26
[#93] Use fixed dark foreground colour for date label in Overview.

jtojnar:
This PR sets a dark foreground colour (default foreground colour of Adwaita theme), so that the date is legible with every theme.

Fixes: #93

### DIFF
--- a/hamster_gtk/resources/css/hamster-gtk.css
+++ b/hamster_gtk/resources/css/hamster-gtk.css
@@ -1,5 +1,6 @@
 #DayRowDateBox {
 	background: #dfdfdf;
+	color: #2e3436;
 }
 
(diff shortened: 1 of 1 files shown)


elbenfreund:
Thanks for the PR.
I tried to stay away from theming specific problems for at least another couple of weeks.
During my initial research I got the impression that providing proper theme support for the ever so fast developing GTK 3 is not trivial and documentation widely spread out.
Whilst I think this solution may still be better than the current status quo it would be nice to investigate best practices on the matter at some point.

elbenfreund:
Result of the IRC discussion: We are hardcoding the background anyway, so we might as well go all the way and do the same to the foreground.

PR #91, opened 2016-09-26
Set transient window for AboutDialog

jtojnar:
Not setting a transient window is discouraged and Gtk will complain.


elbenfreund:
Hey :)
Whilst for general cleanup PRs creating a dummy issue is not always needed and sometimes just creates too much overhead, it would be great if in general we create an issue to link the PR to.
The issue is the best place to explain the actual work at hand/problem while the PR should focus on documenting the implementation/solution. Whilst sometimes a bit tedious it makes documentation and transparency so much easier in the midterm.

PR #92, opened 2016-09-26
Use more general python version for tox

jtojnar:
Use ``python3`` instead of ``python3.4`` in tox environments.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,7 +14,7 @@ commands =
 	make coverage
 
 [testenv:flake8]
-basepython = python3.4
+basepython = python3
 deps =
     flake8==3.0.4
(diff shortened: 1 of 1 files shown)


elbenfreund:
What is the rationale for this PR? In general it seems more sensible information to know *which python version* exactly we pass.

jtojnar:
The reason is I only have Python 3.5 due to which I have to update this file all the time. Python versions should be mostly backwards compatible, so is should really only cause problems if the developer uses much older version than we do.

elbenfreund:
The real solution of cause for this is to use [pyenv](https://github.com/yyuu/pyenv).
But this can be improved further down the road, so I'm ok with this for now.

Issue #95, opened 2016-09-28
Use xvfb for test setup


elbenfreund:
In order to avoid interfering with the developers actual X session an to allow for headless / CI test environments we should evaluate the usage of [xvfb](https://en.wikipedia.org/wiki/Xvfb).

There also seems to exist a [pytest plugin](https://github.com/cgoldberg/xvfbwrapper).

PR #100, opened 2016-09-28
[#95] Add 'xvfb' to our test setup.

elbenfreund:
In order to provide a X server environment even on headless systems as
well as preventing to mess the the calling developers running X session
we uses ``xvfb`` to wrap our tests.
While this adds an editional dependency it is one that only matters for
developers, so it is a viable tradeoff.

Closes: #95

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,17 +67,18 @@ lint:
 	flake8 hamster-dbus tests
 
 test:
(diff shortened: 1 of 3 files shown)


jtojnar:
We might want to add `-a` option to `xvfb-run` in case the display is already occupied.

http://stackoverflow.com/a/25610678/160386

Issue #105, opened 2016-09-29
Disable “click to show more details ...” button when there are no facts


jtojnar:
Clicking “click to show more details ...” button when there are no facts raises an error. It should therefore be disabled.

Other option is hiding the button altogether but that might be more confusing.

PR #107, opened 2016-09-29
[#105] Make details button insensitive when there are no facts

jtojnar:
In the Overview dialogue, if there are no facts and the “click to show more details ...” button is pressed an error is raised. To prevent this, the commit desensitizes the button in those cases.

Closes: #105

### DIFF
--- a/hamster_gtk/overview/dialogs.py
+++ b/hamster_gtk/overview/dialogs.py
@@ -120,9 +120,9 @@ class OverviewDialog(Gtk.Dialog):
         self.totals_panel = widgets.Summary(self._get_highest_totals(self._totals.category, 3))
         self.main_box.pack_start(self.totals_panel, False, False, 0)
 
-        # [FIXME]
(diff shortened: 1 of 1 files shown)


hamster_gtk/overview/dialogs.py
```diff
@@ -120,9 +120,9 @@ def refresh(self):
         self.totals_panel = widgets.Summary(self._get_highest_totals(self._totals.category, 3))
         self.main_box.pack_start(self.totals_panel, False, False, 0)
 
-        # [FIXME]
-        # Only show button if there are facts.
         charts_button = Gtk.Button('click to show more details ...')
+        if len(self._facts) == 0:
```
elbenfreund: should be ``if not self._facts:`` i think.

jtojnar: I will not have time to finish this until tomorrow, feel free to fix it.

Issue #106, opened 2016-09-29
Overview chart not destroyed when changing date range


jtojnar:
When you change the date range in the Overview dialogue while having chart (details) open, the chart will disappear but it will not be destroyed. Two clicks will then be required to show the chart again.

The `self._charts` property should be destroyed on refresh.

PR #108, opened 2016-10-01
[#106] Destroy overview chart when switching date range

jtojnar:
When user changed the date range in the Overview dialogue while having chart (details) open, the chart would disappear but it would not be destroyed. Two clicks will then be required to show the chart again, first to actually destroy it and second to finally display it.

This commit ensures the widget is destroyed whenever the view is refreshed.

Closes: #106

### DIFF
--- a/hamster_gtk/overview/dialogs.py
+++ b/hamster_gtk/overview/dialogs.py
@@ -58,9 +58,11 @@ class OverviewDialog(Gtk.Dialog):
         self.set_default_size(640, 800)
         self.set_titlebar(self.titlebar)
         self.main_box = self.get_content_area()
-        self._daterange = self._get_default_daterange()
-
+        # ``self._charts`` needs to be setup before we assign
+        # ``self._daterange`` as this will trigger ``self.refresh`` which
+        # expects ``self._charts``.
         self._charts = False
+        self._daterange = self._get_default_daterange()
 
         # [FIXME] Should be a property to make sure the signal is emitted
         self._facts = None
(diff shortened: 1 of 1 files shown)


hamster_gtk/overview/dialogs.py
```diff
@@ -109,6 +109,9 @@ def refresh(self):
         self._grouped_facts, self._totals = self._group_facts()
 
         helpers.clear_children(self.main_box)
+        if self._charts:
```
elbenfreund: I am not sure I understand why this works. ``helpers.clear_children(self.main_box)`` should do exactly that the new lines do: call ``destroy`` on all its children. And whenever ``self._charts`` exists its also a child of ``main_box``.
Can you please elaborate?
I can however, confirm that you fix solves the issue.

hamster_gtk/overview/dialogs.py
```diff
@@ -60,8 +62,6 @@ def __init__(self, parent, app, *args, **kwargs):
         self.main_box = self.get_content_area()
         self._daterange = self._get_default_daterange()
 
-        self._charts = False
```
elbenfreund: I am not a huge fan of class attributes. Why does it not suffice to have an instance attribute?

jtojnar: Oh, that is just me not being completely familiar with Python. I just needed for `_charts` to be set before calling refresh (i.e. before `_daterange` property is assigned).

hamster_gtk/overview/dialogs.py
jtojnar: You are right, should be unnecessary to destroy it again.
Even if you destroy the widget, the associated variables will still contain the widget – destroy only cleans the Gtk objects, the corresponding Python objects remain for GC to clear them so we have to set it to False.

elbenfreund: Ah... that makes sense. Thank you.

Issue #109, opened 2016-10-01
Redundant refresh of OverviewDialog


jtojnar:
In `OverviewDialog.__init__` the dialogue is refreshed twice.

1. By assigning to `self._daterange`.
2. By calling `self.refresh()`.

The second refresh should be removed.

PR #110, opened 2016-10-01
[#109] Remove redundant resfresh of OverviewDialog

jtojnar:
In `OverviewDialog.__init__` the dialogue was being refreshed twice.

1. By assigning to `self._daterange`.
2. By calling `self.refresh()`.

This commit removes the explicit refresh call as it is redundant.

Closes: #109

### DIFF
--- a/hamster_gtk/overview/dialogs.py
+++ b/hamster_gtk/overview/dialogs.py
@@ -66,7 +66,6 @@ class OverviewDialog(Gtk.Dialog):
         self._facts = None
         self._grouped_facts = None
 
-        self.refresh()
         self.show_all()
 
     @property


elbenfreund:
This was partially deliberate. Whilst you are absolutly right that it gets called twice and hence is redundant the second call is to *make sure* that a refresh (and hence initial population) of the widget happens independent on what else goes on. That at least was the underlying motivation.

jtojnar:
I do not really understand what else could go on. Assigning to `self._daterange` should already ensure the initial population.

elbenfreund:
It does. My original concern was just: "what if the daterange logic changes"? then we would no longer have a refresh in ``__init___`` and from a "lets refactor daterange" perspective it is not really its responsibility to take care of inits refresh.

jtojnar:
daterange does not care about its refresh, the Overview dialogue does (using setter). Emitting the `daterange-changed` explicitly or even calling `refresh` directly might be more transparent but I don’t find the current approach that inelegant.

Issue #51, opened 2016-07-18
Harmonize signal naming scheme


elbenfreund:
We should be consistent with our signal names. Right now there is ``tracking-stopped`` as well as ``facts_changed``. Choose one style!

PR #112, opened 2016-10-02
Harmonize ``facts_changed`` signal name

jtojnar:
FactListBox emitted ``facts_changed`` signal, even though the SignalHandler recognizes ``facts-changed`` signal. This worked because something translates underscores to dashes.

This signal seems to be the only one using underscores and has now been unified to use dash (Gtk convention).

Closes: #51


elbenfreund:
Is this the only signal that used underscores?

jtojnar:
Did not see any other.

Issue #118, opened 2016-10-04
Rearrange preferences into tabs


jtojnar:
Use GtkStack to rearrange preferences into tabs.

Concept:

![hamster-preferences](https://cloud.githubusercontent.com/assets/705123/19083025/cd2863f6-8a60-11e6-9848-55860ee9caac.png)

elbenfreund:
Please not that [`Gtk.Stack`](http://lazka.github.io/pgi-docs/Gtk-3.0/classes/Stack.html#class-details) is considered best practice for modern GTK3 applications.

jtojnar:
I initially wanted to use `GtkStack` but then I noticed Nautilus uses `GtkNotebook`.

elbenfreund:
I for the life of me can not find the reference that I am sure as hell once had for that, so I can not back this up any further at this point. But  please, before you invest any time lets see what `#gnome-hackers` replies.
If there is no definitive answer, I will prefer `GTK.Stack` and would hate to see any effort of yours go to waste.

PR #122, opened 2016-10-07
[#118] Rearrange Preferences into tabs

jtojnar:
This commit splits Preferences fields into different categories
(Tracking, Storage), each represented by a Gtk.Stack tab.

While it might seem moot to split them with this little fields,
it will be a base for adding other preferences widgets like
activity/category/tag manager.

Closes: #118

### DIFF
--- a/hamster_gtk/preferences/preferences_dialog.py
+++ b/hamster_gtk/preferences/preferences_dialog.py
@@ -31,6 +31,7 @@ from gi.repository import GObject, Gtk
 from hamster_gtk.preferences.widgets import (ComboFileChooser,
                                              HamsterComboBoxText,
                                              HamsterSpinButton,
+                                             PreferencesGrid,
                                              SimpleAdjustment, TimeEntry)
 
 
@@ -45,6 +46,7 @@ class PreferencesDialog(Gtk.Dialog):
             parent (Gtk.Window): Dialog parent.
             app (HamsterGTK): Main app instance. Needed in order to retrieve
                 and manipulate config values.
+            initial (dict): Dictionary of initial config key/values.
         """
         super(PreferencesDialog, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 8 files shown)


jtojnar:
There are two points to address:
- Should the StackSwitcher be placed in dialogue content area or in the header bar?
- The ‘declarative’ `_fields` will not do us much good when we want to add the Category manager as it does not fit the label/widget grid. Should we replace it in favour of:
  - manually creating and attaching widgets
  - replacing the inner `OrderedDict`s with custom class like `GridCreatorWidget(Gtk.Widget)` (encapsulating the current functionality)
  - finally moving to `GtkBuilder`

elbenfreund:
I will not have time to work on hamster today, so here is just some brief feedback.
1. I would place it as the only element of the header bar right now.
2. I am not sure I follow. If we would house the "organize categories etc" stuff here I figured it would look like this (similar to the tracking screen/stack):
- 'tabs' will be separate container widget instances. We should create a custom class that implements the "declarative placement" behavior. So each 'tab' instance has its own `_fields` and works similar to our previous dialog.
- 'Manage categories etc.. ' style 'tabs' will be instances of different container class that does not have this kind of `_fields` related logic.

However, I am not sure that I agree with _legacy hamster_ that 'Preferences' is actually the right place to have 'manage categories/Activities/tags....'. I can see some appeal in this, but right now I rather have a separate dialog for entity management all together. Possibly having a `Gtk.Stack` tab for category, tag etc..

jtojnar:
1. The problem with placing the StackSwitcher to HeaderBar is that modern GTK often places dialogue action buttons into the header bar, something like the following:
   
   actions in header bar
2. I will implement the `DeclarativeGrid` class for now, we can decide about the categories/tags later.

elbenfreund:
For reference: [This link](https://wiki.gnome.org/HowDoI/Dialogs#Headerbars) underlines the use of the header bar for action buttons. Also see the [HIG](https://developer.gnome.org/hig/stable/dialogs.html.en) for more.

If we move the StackSwitcher into the content area, the result will be that action buttons will be "above" the switcher. Which is a bit odd tbh. But then again this may very well be just years of habit with existing UIs.

I do not really have a strong opinion for either solution and as it seems we can revert the decision rather easily at some later point I will go with whatever you prefer for now.

elbenfreund:
After checking with `#gnome-hackers` I got this feedback:

> (15:22:28) mcatanzaro: elbenfreund: Hm, if you have a stack switcher then that does make things more complicated. I wouldn't put any buttons above the stack switcher.
> (15:22:42) mcatanzaro: Maybe to the side? Not sure

jtojnar:
Currently it does look like this which IMHO is not very good.

preferences using stack switcher

The initial implementation using Notebook looked a lot better.

preferences using notebook

Moving the StackSwitcher to HeaderBar is possible but the buttons look like they are only part of the currently selected tab.

preferences using stack switcher in header bar

Possibly we could solve this by saving the values whenever their entry is changed like most GTK apps do. But that might need to be a different PR.

preferences without button

elbenfreund:
I see what you mean. Personally I am not sooo invested in the finer points of the UI as long as future work on that front would not deprecate too much code. I guess it shows in my own contributions. Personally I would give the "StackSwitcher in headerbar" a try. I see you point about the buttons but this should be solveable with a bit of css (different background like in the notebook version) or as you said in a follow up PR that saves changed values right away.

jtojnar:
Even with a different background, it would still be inside the window groove and will look weird. For now it will use Gtk.Notebook and no Gtk.HeaderBar.

Gtk.Notebook, no HeaderBar

It looks the worst with Gtk.HeaderBar.

Gtk.Notebook with HeaderBar

jtojnar:
Apparently if we want to centre the StackSwitcher inside the HeaderBar, it has to be added as a title:

https://github.com/GNOME/gtk/blob/d505a1a/demos/icon-browser/window.ui#L19

We shall use Gtk.Notebook for now.

Issue #25, opened 2016-06-17
Re-enable 'autodocs' for sphinx


elbenfreund:
Initially autodocs failed as they could not import `gi`. Once we have setup proper sitepackages on the CI server this should not be an issue anymore.

PR #128, opened 2016-10-10
[#25] Re-enable autodoc for Sphinx

jtojnar:
Finally managed to tack this. There are actually two problems:
1. Global packages are not added to tox’s environment
2. `_gi.cpython-34m-x86_64-linux-gnu.so` is not recognized as `_gi`.

Both solutions are kinda hacky but after several hours I was not able to come up with anything better.

I am not sure whether to include the second commit. Personally I would delete the files and add them to `.gitignore` as they are generated automatically, but `hamster_gtk.rst` was already committed so I committed the rest too.

The tests are failing due to invalid links (#126).

Relates to: #25

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,6 +28,10 @@ import shlex
 cwd = os.getcwd()
 project_root = os.path.dirname(cwd)
 
+# On Travis python3-gi is installed into folllowing directory, which
+# is not added to path because Travis uses virtualenv.
+sys.path.insert(0, '/usr/lib/python3/dist-packages')
+
 # Insert the project root dir as the first element in the PYTHONPATH.
 # This lets us ensure that the source package is imported, and that its
 # version is used.
@@ -43,7 +47,8 @@ import hamster_gtk
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions = [
-    #'sphinx.ext.autodoc',
+    'sphinx.ext.autodoc',
+    'sphinx.ext.napoleon',
     'sphinx.ext.viewcode',
     'sphinx.ext.todo',
     'sphinx.ext.coverage',
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,9 @@ addons:
     - python-cairo
     - python-gi-cairo
     # - python3-gobject Provided by python3-gi
-    #- python3-gi
-    #- python3-cairo
-    #- python3-gi-cairo
+    - python3-gi
+    - python3-cairo
+    - python3-gi-cairo
 
 virtualenv:
   system_site_packages: true
@@ -38,6 +38,8 @@ install:
   - pip install --upgrade pip
   - pip install -r requirements/test.pip
   - sudo apt-get install --only-upgrade libgtk-3-dev
+  # Trusty only provides modules for Python 3.4, they will, however, work even with newer versions.
+  - sudo ln -s /usr/lib/python3/dist-packages/gi/_gi.cpython-34m-x86_64-linux-gnu.so /usr/lib/python3/dist-packages/gi/_gi.so
 
 before_script:
   - "export DISPLAY=:99.0"
--- a/tox.ini
+++ b/tox.ini
@@ -49,6 +49,7 @@ commands =
 [testenv:docs]
 basepython = python3
 deps = doc8==0.7.0
+sitepackages = True
 commands =
     pip install -r requirements/docs.pip
     make docs BUILDDIR={envtmpdir} SPHINXOPTS={env:SPHINXOPTS_BUILD:''}


jtojnar:
We need to check [`importlib.machinery.EXTENSION_SUFFIXES`](https://docs.python.org/3/library/importlib.html#importlib.machinery.EXTENSION_SUFFIXES). This is how it looks like on my computer: `['.cpython-35m-x86_64-linux-gnu.so', '.abi3.so', '.so']`

Cpython probably is not enabled in the virtualenv.

docs/conf.py
```diff
@@ -28,6 +28,10 @@
 cwd = os.getcwd()
 project_root = os.path.dirname(cwd)
 
+# On Travis python3-gi is installed into folllowing directory, which
+# is not added to path because Travis uses virtualenv.
+sys.path.insert(0, '/usr/lib/python3/dist-packages')
```
elbenfreund, line 33: Do we understand why `system_site_packages: true` does not enable access to the packed installed in the system dir?

jtojnar:
Okay, tox's python has `importlib.machinery.EXTENSION_SUFFIXES` set `['.cpython-35m-x86_64-linux-gnu.so', '.abi3.so', '.so']`, while `python3-gi` package on trusty contains file `_gi.cpython-34m-x86_64-linux-gnu.so`. That means we either need to use python3.4 for building docs or install `python3-gi` with better file names. Or keep making a symlink.

docs/conf.py
jtojnar: Probably because tox detects the global installation incorrectly.

There are multiple python installations:
- system python (`/usr/bin/python3.4`)
- travis’s (`/home/travis/virtualenv/python2.7_with_system_site_packages/bin/python`)
  `sys.path == ['', '/home/travis/virtualenv/python2.7_with_system_site_packages/lib/python2.7', '/home/travis/virtualenv/python2.7_with_system_site_packages/lib/python2.7/plat-x86_64-linux-gnu', '/home/travis/virtualenv/python2.7_with_system_site_packages/lib/python2.7/lib-tk', '/home/travis/virtualenv/python2.7_with_system_site_packages/lib/python2.7/lib-old', '/home/travis/virtualenv/python2.7_with_system_site_packages/lib/python2.7/lib-dynload', '/usr/lib/python2.7', '/usr/lib/python2.7/plat-x86_64-linux-gnu', '/usr/lib/python2.7/lib-tk', '/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages', '/home/travis/virtualenv/python2.7_with_system_site_packages/lib/python2.7/site-packages', '/usr/local/lib/python2.7/site-packages', '/usr/local/lib/python2.7/dist-packages', '/usr/lib/python2.7/dist-packages']`
- tox’s (`/home/travis/build/projecthamster/hamster-gtk/.tox/docs/bin/python`)
  `sys.path == ['', '/home/travis/build/projecthamster/hamster-gtk', '/home/travis/build/projecthamster/hamster-gtk/hamster_gtk', '/home/travis/build/projecthamster/hamster-gtk/.tox/docs/lib/python35.zip', '/home/travis/build/projecthamster/hamster-gtk/.tox/docs/lib/python3.5', '/home/travis/build/projecthamster/hamster-gtk/.tox/docs/lib/python3.5/plat-linux', '/home/travis/build/projecthamster/hamster-gtk/.tox/docs/lib/python3.5/lib-dynload', '/opt/python/3.5.0/lib/python3.5', '/opt/python/3.5.0/lib/python3.5/plat-linux', '/home/travis/build/projecthamster/hamster-gtk/.tox/docs/lib/python3.5/site-packages', '/opt/python/3.5.0/lib/python3.5/site-packages']`
- ?'s (`/opt/python/3.5.0/bin/python`)

tox seems to recognize the last one as global instalation.

jtojnar: When `sitepackages = True` for tox, it runs virtualenv with `--system-site-packages` argument. It runs it [using `sys.executable`](https://github.com/tox-dev/tox/blob/7954ed221f9c9e4fe2fed86dfeb419f1776845ab/tox/venv.py#L400), how is `tox` itself being run?

jtojnar:
``` shell
$ cat `which tox`
#!/home/travis/virtualenv/python2.7_with_system_site_packages/bin/python2.7
[...]
```

That does not help us much.

jtojnar: Virtualenv’s site-packages functionality is defined in [activate_this.py](https://github.com/pypa/virtualenv/blob/263283be88061b965ead9ebff47c37021d7d3f62/virtualenv_embedded/activate_this.py#L18)

jtojnar: So the problem can be caused by the following:
1. Travis's Python creates a virtualenv with correctly detected site package directory.
2. In the created environment, tox creates another virtualenv with the Travis's Python in `sys.executable`, which is actually somehow symlinked to `/opt/python`.

The Travis uses Python 2.7, though, and only some files are symlinked to `/opt/python.` How does the leap from Python 2 to Python 3 actually happen?

elbenfreund:
Thank you @jtojnar for your great work tracking down this mess.
I opted to just run the ``docs`` test under python 2, avoiding the import issue entirely as there should be no cost associated with this solution (#156).

I will close this for now, please feel free to reopen if you feel its needed!

PR #134, opened 2016-10-11
[#133] Remove autodoc generated files

### LINKED ISSUE #133: Remove auto-doc files from repository
Sphinx auto-docs autogenerate files should not be included in the repository. It seems at least `docs/hamster_gtk.rst` has already been added. If so, it should be removed and put on the ignore list.

jtojnar:
This commit removes accidentally added files generated by autodoc. It also adds them to `.gitignore` and excludes them in `MANIFEST.in`.

This also means that developers should not create documentation files starting with `hamster_gtk.` prefix. Trouble when trying `git add` should alert them that it is a bad idea.

Closes: #133


elbenfreund:
In the long term, it may be preferable to move autodoc files to a separate directory.

Issue #136, opened 2016-11-02
Provide a list of 'recent activities' as 'template' for new ongoing facts


elbenfreund:
Similar to ``hamster-shell-extension`` it would be nice if our main tracking screen would show a list of 'recent activities' including a button to use a particular such activity as 'template' for the 'new ongoing fact'.

The usecase is to allow for fast and easy "continue doing XYZ" workflows. Instead of typing the previous activity again such functionality can limit the effort to two to three actions.

Please note that this is a feature in addition / complementing any autocomplete solution.

Details that need feedback:
- [ ] Should the new 'ongoing fact' be *started* right away or just 'pre filled' in the entry field. The latter would require a dedicated 'start tracking' action by the user as an additional failsafe or to allow for amendments.
- [ ] If the new fact is not started right away, focus should be placed on the entry field. Should the field be in a state where the 'pre filled' activity is selected so it can be easily 'overwritten' or would it be preferable if the cursor is placed at the end of the filled in activity string so the user can add tags and descriptions if need be?

elbenfreund:
@jtojnar If you got a minute, I would be gratefull for you opinion on the UX questions...

jtojnar:
I still feel like this should be handled by an extension. Otherwise, one-click option is probably preferable since the goal is probably to allow quick continuation of a fact. Perhaps asking someone who would use the feature would be wiser.

elbenfreund:
Ah, sorry. I forgot about your reservations. This is actually a personal itch for me to scratch. So while I see the appeal to move it to a plugin infrastructure I would like to implement this rather now than postpone until such generic plugin infrastructure is conceived.

PR #138, opened 2016-11-04
Implement autocomplete for 'start tracking'

elbenfreund:
**DO NOT MERGE**

@jtojnar This is my work on autocompletion so far, please have a look if you got a minute and let me know what you think. Once we feel this is the right direction I will try to add some more tests.

This commit adds basic auto completion support to the 'raw_fact entry' of
our ``StartTrackingSreeen``.
For this we make use of ``Gtk.EntryCompletion``. An additional
difficulty is the fact that our raw string may contain 'non completion
relevant' information such as time ranges or descriptions whist the
completion should be limited to ``activity.name``/``activity.category.name``
combinations. By default ``Gtk.EntryCompletion`` substitutes the
entire entry string with the found match. This would mean we loose the
additional information.
- In order to avoid this we first create a 'temporary fact' that contains
all the parsed information present in the entry string so far.
- Then we take any activity/category related information in there to match
against our existing activities.
- Finally we use any matches to feed them back into our 'temporary
fact'.
- If a match was found, a serialized version of the 'tmp fact' gets
placed in the entry field.

Some things are worth noting:
- Due to the fact that we employ ``Fact.create_from_raw_fact`` we apply
our whole 'datetime completion' machinery to the user provided string.
Once the fact gets "written back into the entry" this means the user
provided (potentially partial) information is replaced by the fully
completed datetime information. Whist this was not our original goal it
is actually benefitial as it gives the user clear feedback about what
times will be used. it also fits with the "autocomplete theme".
- For now we provide our own 'serialize fact' helper method. This should
be obsolete once
[LIB-216](https://projecthamster.atlassian.net/browse/LIB-216) has been implemented.

### DIFF
--- a/hamster_gtk/helpers.py
+++ b/hamster_gtk/helpers.py
@@ -19,10 +19,13 @@
 
 """General purpose helper methods."""
 
+from __future__ import absolute_import, unicode_literals
+
 import datetime
+import os.path
 
 import six
-import os.path
+from six import text_type
 
 
 def _u(string):
@@ -117,7 +120,7 @@ def calendar_date_to_datetime(date):
 
 # [FIXME]
 # Remove once hamster-lib is patched
-# This should probablyy be named/limited to: 'read_config_file'.
+# This should probably be named/limited to: 'read_config_file'.
 # The 'fallback' behaviour should live with ``_get_config_from_file``.
 def get_config_instance(fallback_config_instance, app_name, file_name):
     """Patched version of ``hamster-lib`` helper function until it get fixed upstream."""
@@ -135,3 +138,39 @@ def get_config_instance(fallback_config_instance, app_name, file_name):
 def get_resource_path(file_name):
     """Return path to a resource file."""
     return os.path.join(os.path.dirname(__file__), 'resources', file_name)
+
+
+def serialise_activity(activity):
+    """Provide a serialised activity string representation."""
+    if activity.category:
+        activity_string = '{a.name}@{a.category.name}'.format(a=activity)
+    else:
+        activity_string = activity.name
+    return text_type(activity_string)
+
+
+def serialise_fact(fact):
+    """Provide a serialised fact string representation."""
+    # [FIXME] This should be obsolete once LIB-216 is implemented.
+    start_string = ''
+    if fact.start:
+        start_string = fact.start.strftime('%Y-%m-%d %H:%M')
+
+    end_string = ''
+    if fact.end:
+        end_string = ' - ' + fact.end.strftime("%Y-%m-%d %H:%M")
+
+    timestring = start_string + end_string
+    if timestring:
+        timestring += ' '
+
+    category_string = ''
+    if fact.activity.category:
+        category_string = '@{}'.format(fact.activity.category.name)
+
+    label = '{time}{activity}{category}'.format(
+        time=timestring,
+        activity=text_type(fact.activity.name),
+        category=text_type(category_string)
+    )
+    return text_type(label)
--- a/hamster_gtk/misc/widgets/raw_fact_entry.py
+++ b/hamster_gtk/misc/widgets/raw_fact_entry.py
@@ -1,0 +1,127 @@
+# -*- encoding: utf-8 -*-
+
+
+# This file is part of 'hamster-gtk'.
+#
+# 'hamster-gtk' is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# 'hamster-gtk' is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with 'hamster-gtk'.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Widget meant to handle 'raw-fact' strings and provides autocompletion."""
+from __future__ import absolute_import, unicode_literals
+import datetime
+
+from gi.repository import GObject, Gtk
+from orderedset import OrderedSet
+from six import text_type
+from hamster_lib import Category, Fact
+from hamster_gtk.helpers import _u
+from hamster_gtk import helpers
+
+
+class RawFactEntry(Gtk.Entry):
+    """A custom entry widgets that provides ``faw fact`` specific autocompletion behaviour."""
+
+    def __init__(self, controller, *args, **kwargs):
+        """Instantiate the class."""
+        super(RawFactEntry, self).__init__(*args, **kwargs)
+        self._controller = controller
+        self._controller.signal_handler.connect('facts-changed', self._on_facts_changed)
+        self.set_completion(RawFactCompletion(self._controller))
+
+    # Callbacks
+    def _on_facts_changed(self, evtl):
+        """Callback triggered when facts have changed."""
+        self.set_completion(RawFactCompletion(self._controller))
+
+
+class RawFactCompletion(Gtk.EntryCompletion):
+    """
+    Return a completion instance to match 'activity@category' strings.
+
+    Returns:
+        Gtk.EntryCompletion: Completion instance.
+    """
+
+    def __init__(self, controller, *args, **kwargs):
+        """instantiate class."""
+        super(RawFactCompletion, self).__init__(*args, **kwargs)
+        self._controller = controller
+        self.set_model(self._get_store())
+        self.set_text_column(0)
+        self.set_match_func(self._match_anywhere, None)
+        self.connect('match-selected', self._on_match_selected)
+
+    def _get_store(self):
+        store = Gtk.ListStore(GObject.TYPE_STRING, GObject.TYPE_STRING, GObject.TYPE_STRING)
+        for activity in self._get_activities():
+            category = ''
+            if activity.category:
+                category = text_type(activity.category.name)
+            store.append([
+                helpers.serialise_activity(activity), text_type(activity.name), category
+            ])
+        return store
+
+    def _get_activities(self):
+        today = datetime.date.today()
+        recent_activities = [fact.activity for fact in self._controller.facts.get_all(
+            start=today, end=today)]
+        return OrderedSet(recent_activities)
+
+    def _match_anywhere(self, entrystr, iter, data, *args):
+        """
+        Check if the entrysring is a substring of our text_column.
+
+        Args:
+            entrystr (str): The text extracted from the entry so far. Note that key is
+                normalized and case-folded (see GLib.utf8_normalize() and
+                GLib.utf8_casefold()).
+            iter: Iterator position indicating the model row we check against.
+            data: Arbitrary user data.
+
+        Returns:
+            boot: ``True`` if ``entrystring`` is a substring, ``False`` if not.
+
+        Note this is a custom match function, for details on the general generic
+        solution [please see|https://lazka.github.io/pgi-docs/#Gtk-3.0/
+        callbacks.html#Gtk.EntryCompletionMatchFunc].
+        """
+        # 'activity@category' string that we try to match against.
+        modelstr = _u(self.get_model()[iter][0])
+        # Parse the entire entry string to extract all semantic information.
+        fact = Fact.create_from_raw_fact(_u(entrystr))
+        # Make the extracted ``Fact`` available to other callbacks.
+        self._tmp_fact = fact
+        # If the parsed string seems to contain any activity related
+        # information, encode it in its canonical form.
+        if fact.activity:
+            activity_string = helpers.serialise_activity(fact.activity)
+        return activity_string in modelstr
+
+    def _on_match_selected(self, model, iter):
+        """Callback to be executed once a match is selected by the user."""
+        activity_name = _u(model[iter][1])
+        category_name = _u(model[iter][2])
+        fact = self._tmp_fact
+        # We complete ``self._tmp_fact`` with activity/category information
+        # taken from the matching model row.
+        fact.activity.name = activity_name
+        if category_name:
+            fact.activity.category = Category(category_name)
+        else:
+            fact.activity.category = None
+        # Put a string that contains all available and completed information
+        # in the text entry.
+        self.get_entry().set_text(helpers.serialise_fact(fact))
+        self.get_entry().set_position(-1)
+        return True
--- a/hamster_gtk/tracking/screens.py
+++ b/hamster_gtk/tracking/screens.py
@@ -28,6 +28,7 @@ from gettext import gettext as _
 from gi.repository import GObject, Gtk
 from hamster_lib import Fact
 
+from hamster_gtk.misc.widgets import RawFactEntry
 import hamster_gtk.helpers as helpers
 from hamster_gtk.helpers import _u
 
@@ -35,17 +36,17 @@ from hamster_gtk.helpers import _u
 class TrackingScreen(Gtk.Stack):
     """Main container for the tracking screen."""
 
-    def __init__(self, app, *args, **kwargs):
+    def __init__(self, controller, *args, **kwargs):
         """Setup widget."""
-        super(TrackingScreen, self).__init__()
-        self.app = app
+        super(TrackingScreen, self).__init__(*args, **kwargs)
+        self._controller = controller
 
         self.main_window = helpers.get_parent_window(self)
         self.set_transition_type(Gtk.StackTransitionType.SLIDE_UP)
         self.set_transition_duration(1000)
-        self.current_fact_view = CurrentFactBox(self.app.controller)
+        self.current_fact_view = CurrentFactBox(self._controller)
         self.current_fact_view.connect('tracking-stopped', self.update)
-        self.start_tracking_view = StartTrackingBox(self.app.controller)
+        self.start_tracking_view = StartTrackingBox(self._controller)
         self.start_tracking_view.connect('tracking-started', self.update)
         self.add_titled(self.start_tracking_view, 'start tracking', _("Start Tracking"))
         self.add_titled(self.current_fact_view, 'ongoing fact', _("Show Ongoing Fact"))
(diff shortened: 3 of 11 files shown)


hamster_gtk/helpers.py
```diff
@@ -135,3 +138,39 @@ def get_config_instance(fallback_config_instance, app_name, file_name):
 def get_resource_path(file_name):
     """Return path to a resource file."""
     return os.path.join(os.path.dirname(__file__), 'resources', file_name)
+
+
+def serialize_activity(activity):
+    """Provide a serialised activity string representation."""
```
jtojnar, line 144: In the comments you use ‘serialised’ while the rest of the app uses American spelling. This should probably be unified.

hamster_gtk/helpers.py
```diff
@@ -135,3 +138,39 @@ def get_config_instance(fallback_config_instance, app_name, file_name):
...
+    if activity.category:
+        activity_string = '{a.name}@{a.category.name}'.format(a=activity)
+    else:
+        activity_string = activity.name
+    return activity_string
+
+
+def serialize_fact(fact):
+    """Provide a serialised fact string representation."""
```
jtojnar, line 153: In the comments you use ‘serialised’ while the rest of the app uses American spelling. This should probably be unified.

hamster_gtk/tracking/screens.py
```diff
@@ -166,7 +168,15 @@ def __init__(self, controller, *args, **kwargs):
         """Setup widget."""
         super(StartTrackingBox, self).__init__(orientation=Gtk.Orientation.VERTICAL,
                                                spacing=10, *args, **kwargs)
+        # Parsed entry string will store (potentially incomplete) ``Fact``
+        # instances in here so they can be refered to by ``ob_match_selected``.
```
jtojnar: Misspelling ‘refered’

hamster_gtk/tracking/screens.py
```diff
@@ -166,7 +168,15 @@ def __init__(self, controller, *args, **kwargs):
...
+        # It's main use is to store information that is part of the user
```
jtojnar: Misspelling ‘It’s’

hamster_gtk/tracking/screens.py
```diff
@@ -178,13 +188,105 @@ def __init__(self, controller, *args, **kwargs):
...
+        Return a completion instance to match 'activity@category' strings.
+
+        Returns:
+            Gtk.EntryCompletion: Completion instance.
+        """
+        def remove_duplicates(list):
+            """
+            Quick and dirty function to remove duplicate list items.
+
+            The first item encounterd is keept, the following thrown away.
```
jtojnar: Misspelling ‘encounterd’ and ‘keept’

hamster_gtk/tracking/screens.py
```diff
@@ -178,13 +188,105 @@ def __init__(self, controller, *args, **kwargs):
...
+
+            We can not simply turn this into a set as we would loose the
```
jtojnar: Misspelling ‘loose’

hamster_gtk/tracking/screens.py
```diff
@@ -178,13 +188,105 @@ def __init__(self, controller, *args, **kwargs):
...
+            ordering.  This clearly is highly inefficiennt and the only reason
```
jtojnar: Two spaces between ‘ordering.’ and ‘This’; misspelling ‘inefficiennt’

hamster_gtk/tracking/screens.py
```diff
@@ -178,13 +188,105 @@ def __init__(self, controller, *args, **kwargs):
...
+            we do this for now is the very limited size of the input list.
+
+            For one possible way to do this see: ``https://code.activestate.com/recipes/576694/``.
```
jtojnar: There is an efficient OrderedSet implementation on PyPi: https://pypi.python.org/pypi/orderedset

hamster_gtk/tracking/screens.py
```diff
@@ -178,13 +188,105 @@ def __init__(self, controller, *args, **kwargs):
...
+            """
+            result = []
+            for item in list:
+                if item not in result:
+                    result.append(item)
+            return result
+
+        def match_anywhere(completion, entrystr, iter, data):
+            """
+            Check if the entrysring is a substring of our text_column.
```
jtojnar: Misspelling ‘entrysring’

hamster_gtk/tracking/screens.py
```diff
@@ -178,13 +188,105 @@ def __init__(self, controller, *args, **kwargs):
...
+
+            Args:
+                completion (Gtk.EntryCompletion): The completion instance.
+                entrystr (str): The text extracted from the entry so far.
+                iter: Iterator position indicating the model row we check against.
+                data: ???.
```
jtojnar: This is the `None` passed to the `set_match_func` method: https://lazka.github.io/pgi-docs/#Gtk-3.0/callbacks.html#Gtk.EntryCompletionMatchFunc

jtojnar (changes requested):

jtojnar:
1. The completion does not work for category-less activities:
   
   ```
   Traceback (most recent call last):
     File "~/projects/hamster-gtk/hamster_gtk/tracking/screens.py", line 260, in on_match_selected
       fact.activity.category = Category(category_name)
     File "~/.virtualenvs/hamster-gtk/lib/python3.5/site-packages/hamster_lib/objects.py", line 51, in __init__
       self.name = name
     File "~/.virtualenvs/hamster-gtk/lib/python3.5/site-packages/hamster_lib/objects.py", line 61, in name
       raise ValueError(_("You need to specify a name."))
   ValueError: You need to specify a name.
   ```
2. The RawFactWidget should be factored out to a separate class which will encapsulate the completion. This will also allow to use it for both creating and editing the facts.
3. What is the `tracking-canceled` event for? Why is it part of this PR?
4. Should this only suggest the activities from today? I do some activities only once a week and I would still like them to be suggested. Not sure where to draw the line though – one can as well argue for once a year and the suggestion would be even more useful there as the user won’t probably remember the exact name.

jtojnar:
Entering space also produces an error.

elbenfreund:
what do you mean with "entering space"?

jtojnar:
Entering space characters using space bar.

elbenfreund:
Do you mean the error dialog that appears or do you get an unhandled exception? What is your expected behaviour?

jtojnar:
Unhandled error:

```
ValueError: You need to specify a name.
Traceback (most recent call last):
  File "~/hamster-gtk/hamster_gtk/tracking/screens.py", line 243, in match_anywhere
    fact = Fact.create_from_raw_fact(entrystr)
  File "~/hamster-gtk/lib/python3.5/site-packages/hamster_lib/objects.py", line 478, in create_from_raw_fact
    activity = Activity(activity_name, category=category)
  File "~/hamster-gtk/lib/python3.5/site-packages/hamster_lib/objects.py", line 147, in __init__
    self.name = name
  File "~/.virtualenvs/hamster-gtk/lib/python3.5/site-packages/hamster_lib/objects.py", line 159, in name
    raise ValueError(_("You need to specify a name."))
```

Expected behaviour would be something like this:

``` python
if entrystr.strip() == '':
    return True
```

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,127 @@
...
+class RawFactCompletion(Gtk.EntryCompletion):
+    """
+    Return a completion instance to match 'activity@category' strings.
+
+    Returns:
+        Gtk.EntryCompletion: Completion instance.
+    """
+
+    def __init__(self, controller, *args, **kwargs):
+        """instantiate class."""
```
jtojnar, line 56: In [`RawFactEntry.__init__`](https://github.com/projecthamster/hamster-gtk/pull/138/commits/b7e7d875f60314be495f0bb9d2b1b1da192b5c6d#diff-f5e2db386bd89261e3d69c4dfeda3181R35) you use ‘Instantiate the class.’ while here you just use ‘instantiate class.’ It should probably be unified, at least in the same file.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,127 @@
...
+        """Instantiate the class."""
+        super(RawFactEntry, self).__init__(*args, **kwargs)
+        self._controller = controller
+        self._controller.signal_handler.connect('facts-changed', self._on_facts_changed)
+        self.set_completion(RawFactCompletion(self._controller))
+
+    # Callbacks
+    def _on_facts_changed(self, evtl):
+        """Callback triggered when facts have changed."""
+        self.set_completion(RawFactCompletion(self._controller))
```
jtojnar, line 44: Rather than creating a new completion each time, we should probably just propagate the event to the completion and let it update its own model. Or since we are already passing the controller to the completion, the completion could subscribe the `facts-changed` signal on its own.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def _get_activities(self):
+        today = datetime.date.today()
+        recent_activities = [fact.activity for fact in self._controller.facts.get_all(
+            start=today, end=today)]
+        return OrderedSet(recent_activities)
+
+    def _match_anywhere(self, entrystr, iter, data, *args):
+        """
+        Check if the entrysring is a substring of our text_column.
```
jtojnar, line 83: Misspelling entrysring

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,127 @@
...
+
+        Args:
+            entrystr (str): The text extracted from the entry so far. Note that key is
+                normalized and case-folded (see GLib.utf8_normalize() and
+                GLib.utf8_casefold()).
+            iter: Iterator position indicating the model row we check against.
+            data: Arbitrary user data.
+
+        Returns:
+            boot: ``True`` if ``entrystring`` is a substring, ``False`` if not.
```
jtojnar, line 93: Misspelling boot

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,127 @@
...
+            ])
+        return store
+
+    def _get_activities(self):
+        today = datetime.date.today()
+        recent_activities = [fact.activity for fact in self._controller.facts.get_all(
+            start=today, end=today)]
+        return OrderedSet(recent_activities)
+
+    def _match_anywhere(self, entrystr, iter, data, *args):
```
jtojnar, line 81: Where does the `args` argument come from?

https://lazka.github.io/pgi-docs/#Gtk-3.0/callbacks.html#Gtk.EntryCompletionMatchFunc has no mention of it.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,127 @@
...
+                normalized and case-folded (see GLib.utf8_normalize() and
+                GLib.utf8_casefold()).
+            iter: Iterator position indicating the model row we check against.
+            data: Arbitrary user data.
+
+        Returns:
+            boot: ``True`` if ``entrystring`` is a substring, ``False`` if not.
+
+        Note this is a custom match function, for details on the general generic
+        solution [please see|https://lazka.github.io/pgi-docs/#Gtk-3.0/
```
jtojnar, line 96: Links should not really be broken like this. Also what is the `[link|url]` syntax?

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,127 @@
...
+        callbacks.html#Gtk.EntryCompletionMatchFunc].
+        """
+        # 'activity@category' string that we try to match against.
+        modelstr = _u(self.get_model()[iter][0])
+        # Parse the entire entry string to extract all semantic information.
+        fact = Fact.create_from_raw_fact(_u(entrystr))
```
jtojnar, line 102: Creating the fact for each activity in the model is not really efficient. Maybe we could subscribe for entry changes and update the fact then (and only accessing `self._fact` here. Or even `self._activity_string`.

hamster_gtk/tracking/screens.py
```diff
@@ -59,7 +60,7 @@ def update(self, evt=None):
         This depends on wether there exists an *ongoing fact* or not.
```
jtojnar, line 60: Misspelling wether

tests/misc/conftest.py
```diff
@@ -62,3 +66,41 @@ def monthrange_parametrized(request):
...
+    which do not.
+    """
+    model = Gtk.ListStore(GObject.TYPE_STRING, GObject.TYPE_STRING, GObject.TYPE_STRING)
+    model.append(['foo@bar', 'foo', 'bar'])
+    return model
+
+
+@pytest.fixture
+def activity_model(request, activity):
+    """Returnd a ``ListStore`` instance with a generic ``Activity`` as it's only row."""
```
jtojnar, line 99: Misspelling Returnd and it's

jtojnar (changes requested):

jtojnar:
The current code does not work for me. I had to create `hamster_gtk/misc/widgets/__init__.py` re-exporting `RawFactEntry` and add re-export `RawFactEntry` from `hamster_gtk/misc/__init__.py` before I managed to start the application.

I also had to install `OrderedSet` package, it should probably be added to `setup.py`.

Even then, entering any text into the field causes an error

```
(hamster-gtk:30195): Gtk-WARNING **: gtktreemodel.c:645: Invalid path test passed to gtk_tree_path_new_from_string
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/gi/overrides/Gtk.py", line 1162, in __new__
    return TreePath.new_from_string(path)
TypeError: constructor returned NULL

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "~/projects/hamster-gtk/hamster_gtk/misc/widgets/raw_fact_entry.py", line 100, in _match_anywhere
    modelstr = _u(self.get_model()[iter][0])
  File "/usr/lib/python3.5/site-packages/gi/overrides/Gtk.py", line 801, in __getitem__
    aiter = self._getiter(key)
  File "/usr/lib/python3.5/site-packages/gi/overrides/Gtk.py", line 789, in _getiter
    aiter = self.get_iter(key)
  File "/usr/lib/python3.5/site-packages/gi/overrides/Gtk.py", line 823, in get_iter
    path = self._coerce_path(path)
  File "/usr/lib/python3.5/site-packages/gi/overrides/Gtk.py", line 798, in _coerce_path
    return TreePath(path)
  File "/usr/lib/python3.5/site-packages/gi/overrides/Gtk.py", line 1164, in __new__
    raise TypeError("could not parse subscript '%s' as a tree path" % path)
TypeError: could not parse subscript 'test' as a tree path
```

**EDIT:** Actually, it is caused by a missing parameter in a `_match_anywhere` method. See https://github.com/projecthamster/hamster-gtk/pull/138#discussion_r87098991

hamster_gtk/misc/widgets/raw_fact_entry.py
jtojnar: Actually you are missing the `completion` parameter (see the reference). It should be like this:

``` python
def _match_anywhere(self, completion, entrystr, iter, data):
```

It is the cause of the exception I mention in the comment bellow.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,127 @@
...
+        # Make the extracted ``Fact`` available to other callbacks.
+        self._tmp_fact = fact
+        # If the parsed string seems to contain any activity related
+        # information, encode it in its canonical form.
+        if fact.activity:
+            activity_string = helpers.serialise_activity(fact.activity)
+        return activity_string in modelstr
+
+    def _on_match_selected(self, model, iter):
```
jtojnar, line 111: Similarly to https://github.com/projecthamster/hamster-gtk/pull/138#discussion_r87098991 there should be `completion` parameter (see https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/EntryCompletion.html#Gtk.EntryCompletion.signals.match_selected). Otherwise this error is raised:

``` python
TypeError: _on_match_selected() takes 3 positional arguments but 4 were given
```

jtojnar (changes requested):

hamster_gtk/misc/widgets/raw_fact_entry.py
elbenfreund: strange. our signature provides 3 parameter spots just as in the docs you referenced. The only difference is the naming of the first one (`self` vs `entry_completion`) which the code should not care about.

jtojnar: I guess, when passing a method as an argument, the object is bound to `self`, i.e. `obj.method(*args)` is actually functionally equivalent to `lambda *args: obj.method(*args)` – `obj` is captured in a closure, similarly to writing `lambda *args: getattr(obj, 'method')(*args)`

hamster_gtk/misc/widgets/raw_fact_entry.py
elbenfreund: I agree. However, I would have to look into the exact way signals are triggered/handled here in order to prevent race conditions between processing `text-changed` and `match_anywhere`.
In case this turns out to be non trivial it may be possible to take it as it is and open an enhancement issue instead...

jtojnar: Other way is caching the fact string in a property – something like this:

``` python
cache_key = self._activity_string_cache[0]
if cache_key != entrystr:
        fact = Fact.create_from_raw_fact(_u(entrystr))
        if fact.activity:
            activity_string = helpers.serialise_activity(fact.activity)
        self._activity_string_cache = (entrystr, activity_string)
activity_string = self._activity_string_cache[1]
```

elbenfreund:
After some deliberation I think this approach is not all that viable. Whilst it can be made work 'sort of' it its gets hackier by the minute if one tries to have seperate activity.name and activity.category.name autocompletion let alone tags. I will cose this for now and try something else instead.

For reference: Please note that the branch name actually refers to the wrong issue (#136) which itelf is actually not the best fit to begin with.

PR #141, opened 2016-11-08
Show only develop build status in badge

jtojnar:
Change Travis badge to show the build status of develop branch only. Unless we do this, pushes to WIP branches (e.g. `requires-io-develop`) can make it look like our software is broken.


elbenfreund:
Thanks for pointing  this one out. I think the badge should point to `master` though, as this is the authoritative 'latest stable version'.

jtojnar:
I still consider the stable branch something superfluous and annoying. But I guess the badge should be always green anyway so it does not matter.

PR #144, opened 2016-11-11
[#147] Allow choosing export type

jtojnar:
As ``hamster-lib`` already supports multiple export backends, this commit only had to add an interface making it accessible.

Currently, CSV, iCalendar and XML are supported by ``hamster-lib``.

The unit tests will have to be added before merging.

Closes: #145 #147

### DIFF
--- a/hamster_gtk/overview/dialogs/__init__.py
+++ b/hamster_gtk/overview/dialogs/__init__.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+
+# This file is part of 'hamster-gtk'.
+#
+# 'hamster-gtk' is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# 'hamster-gtk' is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with 'hamster-gtk'.  If not, see <http://www.gnu.org/licenses/>.
+
+"""This module provides an overview dialog allowing access to all facts, and an export dialog."""
+
+from .export_dialog import ExportDialog  # NOQA
+from .overview_dialog import OverviewDialog, Totals  # NOQA
--- a/hamster_gtk/overview/dialogs/export_dialog.py
+++ b/hamster_gtk/overview/dialogs/export_dialog.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+
+# This file is part of 'hamster-gtk'.
+#
+# 'hamster-gtk' is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# 'hamster-gtk' is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with 'hamster-gtk'.  If not, see <http://www.gnu.org/licenses/>.
+
+
+"""This Module provides the ``ExportDialog`` class to choose file to be exported."""
+
+from __future__ import absolute_import
+
+import os.path
+from gettext import gettext as _
+
+from gi.repository import Gtk
+
+from hamster_gtk.preferences.widgets import PreferencesGrid
+
+
+class ExportDialog(Gtk.FileChooserDialog):
+    """Dialog used for exporting."""
+
+    def __init__(self, parent):
+        """
+        Initialize export dialog.
+
+        Args:
+            parent (Gtk.Window): Parent window for the dialog.
+        """
+        super(ExportDialog, self).__init__(_("Please choose where to export to"), parent,
+            Gtk.FileChooserAction.SAVE, (Gtk.STOCK_CANCEL, Gtk.ResponseType.CANCEL,
+                                         Gtk.STOCK_SAVE, Gtk.ResponseType.OK))
+
+        self._export_format_chooser = self._get_export_format_chooser()
+        self._export_format_chooser.connect('changed', self._on_export_format_changed)
+
+        export_options = PreferencesGrid(
+            {'format': (_("Export _format:"), self._export_format_chooser)})
+        export_options.show_all()
+        self.set_extra_widget(export_options)
+
+        self.set_current_name(_("hamster_export"))
+        self._export_format_chooser.set_active_id('tsv')
+
+    def _get_export_format_chooser(self):
+        chooser = Gtk.ComboBoxText()
+        chooser.append('tsv', _("TSV"))
+        chooser.append('ical', _("iCal"))
+        chooser.append('xml', _("XML"))
+        return chooser
+
+    def _on_export_format_changed(self, combobox):
+        """
+        Change file extension of the selected file to the one that was chosen.
+
+        Args:
+            combobox (Gtk.ComboBoxText): Combo box that was changed.
+        """
+        new_ext = self.get_export_format()
+        (name, ext) = os.path.splitext(self.get_current_name())
+        self.set_current_name('{}.{}'.format(name, new_ext))
+
+    def get_export_format(self):
+        """
+        Return currently selected export format.
+
+        Returns:
+            text_type: Currently selected export format.
+        """
+        return self._export_format_chooser.get_active_id()
--- a/hamster_gtk/overview/dialogs/export_dialog.py
+++ b/hamster_gtk/overview/dialogs/export_dialog.py
@@ -25,7 +25,7 @@ from gettext import gettext as _
 
 from gi.repository import Gtk
 
-from hamster_gtk.preferences.widgets import PreferencesGrid
+from hamster_gtk.misc.widgets import LabelledWidgetsGrid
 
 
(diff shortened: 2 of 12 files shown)


jtojnar:
User Interface

I had to create a widget for the selector. There is supposed to be [add_choice](https://lazka.github.io/pgi-docs/#Gtk-3.0/interfaces/FileChooser.html#Gtk.FileChooser.add_choice) method but it segfaults for me.

elbenfreund:
Thanks for the info. `add_choice` is >3.22 anyway. Seeing that we target >3.10 not using it is a good thing :)
We probably should create an issue though so we may switch to `add_choice` in some later iteration

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+
+from __future__ import absolute_import
+
+import os.path
+from enum import Enum
+from gettext import gettext as _
+
+from gi.repository import Gtk
+
+ExportType = Enum('ExportType', ('CSV', 'ICAL', 'XML'))
```
elbenfreund: Why do you use a 'Enum' here instead of just a dict?
Btw: Future iteration should be able to avoid this altogether as the backend will provide a registry for exactly this purpose. [LIB-232](https://projecthamster.atlassian.net/browse/LIB-232)

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class ExportDialog(Gtk.FileChooserDialog):
+    """Dialog used for exporting."""
+
+    def __init__(self, parent):
+        """
+        Initialize headerbar.
```
elbenfreund: 'headerbar' should be 'dialog'

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+
+        Args:
+            parent (Gtk.Window): Parent window for the dialog.
+        """
+        super(ExportDialog, self).__init__(_("Please Choose where to export to"), parent,
+            Gtk.FileChooserAction.SAVE, (Gtk.STOCK_CANCEL, Gtk.ResponseType.CANCEL,
+                                         Gtk.STOCK_SAVE, Gtk.ResponseType.OK))
+
+        self._export_type_chooser = Gtk.ComboBoxText()
```
elbenfreund: It would be better to wrap all the `type_chooser` related code into a private method like `self._get_type_choose` that returns the deals with creating label and dropdown and just returns the grid.

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+        self._export_type_chooser.append(str(ExportType.CSV), _("CSV"))
```
elbenfreund: Export types do not need localization or?

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+        self._export_type_chooser.append(str(ExportType.ICAL), _("iCal"))
+        self._export_type_chooser.append(str(ExportType.XML), _("XML"))
+        self._export_type_chooser.connect('changed', self._export_type_changed)
```
elbenfreund: callback should be called `_on_export_type_changed`.

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+
+        export_type_label = Gtk.Label(_("Export as file _type:"))
```
elbenfreund: I would prefer `Export as`. I think 'file type' is misleading. They are all 'text files' its just semantics of its content that are different. Alternatively I could live with 'Export type...'.

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+        export_type_label.set_use_underline(True)
+        export_type_label.set_mnemonic_widget(self._export_type_chooser)
+
+        export_options = Gtk.Grid()
+        export_options.attach(export_type_label, 0, 0, 1, 1)
+        export_options.attach_next_to(self._export_type_chooser, export_type_label,
+                                      Gtk.PositionType.RIGHT, 1, 1)
+        export_options.show_all()
```
elbenfreund: Wouldn't it be better to call `self.show_all`? This should traverse down to the combobox as well.

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,102 @@
...
+        self.set_extra_widget(export_options)
+
+        self.set_current_name(_("hamster_export"))
+        self._export_type_chooser.set_active_id(str(ExportType.CSV))
+
+    def _export_type_changed(self, combobox):
```
elbenfreund: should be named _on_export_type_changed

hamster_gtk/overview/dialogs/export_dialog.py
jtojnar: I suppose some locales might want to translate it.

hamster_gtk/overview/dialogs/export_dialog.py
jtojnar: There might be some intentionally hidden widgets in the dialog. I would rather show the narrowest set of widgets possible.

hamster_gtk/overview/dialogs/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with 'hamster-gtk'.  If not, see <http://www.gnu.org/licenses/>.
+
+"""This module provides an overview dialog allowing access to all facts, and an export dialog."""
+
+from .export_dialog import ExportDialog, export_writers  # NOQA
```
elbenfreund: Do we really want to make ``export_writers`` public?
wouldn't it suffice to put it somewhere around [here](https://github.com/projecthamster/hamster-gtk/pull/144/files#diff-1e116303a380135b42893c60abe25a72R255)?

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,84 @@
...
+
+import os.path
+from gettext import gettext as _
+
+from gi.repository import Gtk
+from hamster_lib import reports
+
+from hamster_gtk.preferences.widgets import PreferencesGrid
+
+export_writers = {'csv': reports.TSVWriter, 'ical': reports.ICALWriter, 'xml': reports.XMLWriter}
```
elbenfreund: ``csv`` should be ``tsv``

hamster_gtk/overview/dialogs/export_dialog.py
```diff
@@ -0,0 +1,84 @@
...
+            parent (Gtk.Window): Parent window for the dialog.
+        """
+        super(ExportDialog, self).__init__(_("Please choose where to export to"), parent,
+            Gtk.FileChooserAction.SAVE, (Gtk.STOCK_CANCEL, Gtk.ResponseType.CANCEL,
+                                         Gtk.STOCK_SAVE, Gtk.ResponseType.OK))
+
+        self._export_format_chooser = self._get_export_format_chooser()
+        self._export_format_chooser.connect('changed', self._on_export_format_changed)
+
+        export_options = PreferencesGrid(
```
elbenfreund: This is really messy. Either ``PreferencesGrid`` needs to be renamed/relocated to reflect its more general usage or it should not be used here.

hamster_gtk/overview/widgets/misc.py
```diff
@@ -108,14 +109,12 @@ def _on_export_button_clicked(self, button):
...
-            Gtk.FileChooserAction.SAVE, (Gtk.STOCK_CANCEL, Gtk.ResponseType.CANCEL,
-                                         Gtk.STOCK_SAVE, Gtk.ResponseType.OK))
-        dialog.set_current_name('{}.csv'.format(_("hamster_export")))
+        dialog = ExportDialog(parent)
         response = dialog.run()
         if response == Gtk.ResponseType.OK:
             target_path = dialog.get_filename()
-            parent._export_facts(target_path)
+            target_format = dialog.get_export_format()
+            parent._export_facts(target_format, target_path)
```
elbenfreund: L116 and L117 could probably be merged into one...

tests/overview/test_dialogs.py
```diff
@@ -38,9 +39,9 @@ def test__export_facts(self, overview_dialog, tmpdir, mocker):
         With all its mocks this test is not the best one imageinable, but as
         the method will change rapidly soon this does for now.
         """
-        writer = mocker.patch('hamster_gtk.overview.dialogs.reports.TSVWriter')
+        writer = mocker.patch('hamster_gtk.overview.dialogs.overview_dialog.reports.TSVWriter')
         overview_dialog._get_facts = mocker.MagicMock(return_value={})
-        result = overview_dialog._export_facts(tmpdir.strpath)
+        result = overview_dialog._export_facts('csv', tmpdir.strpath)
```
elbenfreund: this test should probably be parametrized over all supported 'export strings'.

hamster_gtk/overview/widgets/misc.py
jtojnar: Should that also include line 115?

elbenfreund: yes

hamster_gtk/overview/dialogs/export_dialog.py
jtojnar: Should the extension also be `tsv`? It was `csv` previously: https://github.com/projecthamster/hamster-gtk/pull/144/files#diff-3d1746373db586ae8ad2273cc902a5cbL114

elbenfreund: yes please.

Issue #148, opened 2016-11-15
Provide autocompletion when entering 'raw fact' strings.


elbenfreund:
Similar to legacy hamster it would be nice if all entry widgets that deal with 'raw fact' strings would provide autocompletion at least for ``Activity.name`` and ``Activity.category.name``. Supporting tags as well would be a major bonus.

The expected workflow would be:
- user starts entering a raw fact
- when 'reaching' the segment where activity.name or category is typed in, a list of suggestions pops up
- user can select any given entry which then replaces the activity/category typed so far. or user just continues to type without any selection.

Optional bonuses:
- revisiting a already specified segment (moving the cursor back to 'foo' if finsished typing 'foo@bar') will re-show any viable suggestions.
- if raw fact string is empty, pushing the 'down arrow key' shows the last X activities.

The main difficulty here is to determine 'what segment' of the raw entry string the user 'wants to specify' at any given point. We will have to come up with a way to inspect the string and make a guess about what information the user is currently trying to 'encode' (timeinfo, activity, category, tags or description).
Based on that guess we can then launch our autocompletion machinery.

PR #150, opened 2016-11-15
[#148] Autocomple for 'Raw Fact' strings

elbenfreund:
This PR changes the way raw fact strings are handled by our GUI.
Instead of using a generic ``Gtk.Entry`` we introduce a new ``RawFactEntry`` class to handle the specific requirements of 'raw fact' strings. 
Autocompletion is provided by a customized ``EntryCompletion`` class called ``RawFactCompletion``.
Both custom classes are to be used in conjunction.

The basic approach to the problem at hand is as follows.

1. ``RawFactEntry._on_changed`` decomposes the current raw fact string and tries to map its substrings to activity, category, tags, etc... segments. This is done with the help of regex matching.
2. If the entry text could be mapped at all, the entry uses the current cursor position to make a guess what segment the user is currently trying to specify.
3. If the currently edited segment is one that we want to provided autocompletion for, the appropiate model is set on the completion instance and the completion machinery is launched.
4. If the user selects a completion suggestion, the original segment text gets replaced by the completion text and the cursor placed at the segment end.

There are quite some subtleties involved as the result of our non standard matching/autocomplete requirements applying only to substrings of the entry (for more details see the commit msg and code annotations) but this is the gist of it.

One central cornerstone of this solution is our new 'raw fact' regex. This entire approach depends on our ability to come up with a regex that is able to capture the semantics of our raw fact string sufficiently well.
Please not that this does not include any validation at all.
*If* this works out, we may even have a good candidate to refactor ``hamster_lib.Fact.create_from_raw_fact`` to use this regex based solution instead of its current rather hacky one.

Closes: #148


jtojnar:
At a glance, it looks reasonable. I am mostly concerned about the `decompose_raw_fact_string` helper.
1. Descriptions cannot contain at-signs or hashes – `foo@bar, @baz` or `foo@bar, #baz` do not match anything.
2. Tags can contain at-signs, though – `foo #@bar` does match.
3. Relative times are not matched.
4. Category without activity can be matched.
5. The function returns the matches including separators and spaces.

elbenfreund:
You are totally right to be concerned about `decompose_raw_fact_string`. This clearly carries a lot of weight and is hard to get right.
What I did here is just a first and relatively restrictive approximation. By composing the regex in such a modular way we should be able to experiment quite easily to hopefully find something that works nicely.
The challenge is to create a regex that matches valid strings, but also is powerfull and flexible enough to have a good change of matching incomplete strings sufficiently well.
The legacy parsing has its own issues and by now I would be willing to break compatibility if that would be really needed for a clear, concise and transparent raw fact string semantic (for example, one could consider using `;` as a description seperator instead.

Now on to you first feedback.
1. Yes, and I agree that it would be good to allow as many characters in the description (including `@` and `#` as possible. I will add a commit that will give this a try. There is always the risk of side effects but this is something only experience will show.
2. That is deliberate: as I said, I try to limit the char-set as little as possible wherever possible.
3. Yes, thats something still to be done. Thanks for the reminder.
4. That is intensional. As this is not a `create from raw fact` but rather a 'get me all the information contained in the string'  function this is exactly the way we want it to behave.
5. Yes. I wanted to make sure that the function makes as little preassumptions about the way it is used as possible. Right now `RawFactEntry` provides its own prefix removal and alllows easy access to the cleaned up segment text via `RawFactEntry.get_segment_text`. If this a recurring issue, prefix removal could be outsourced to a seperate helper method that provides cleaned up strings.

hamster_gtk/helpers.py
```diff
@@ -137,3 +138,62 @@ def get_config_instance(fallback_config_instance, app_name, file_name):
...
     return os.path.join(os.path.dirname(__file__), 'resources', file_name)
+
+
+def decompose_raw_fact_string(text, raw=False):
+    """
+    Try to match a given string with modular regex groups.
+
+    Args:
+        text (text_type): String to be analysed.
+        raw (bool): If True, return the raw match instance, if ``False`` return
```
jtojnar: `False` is backticked here, while `True` is not.

README.rst
```diff
@@ -33,7 +33,7 @@ To Run the Testsuite
 First Steps
 ------------
 * Install dependencies (on debian if using virtualenvwrapper):
-  ``apt-get install virtualenvwrapper python-gi gir1.2-gtk-3.0 libglib2.0-dev
+  install virtualenvwrapper python-gi gir1.2-gtk-3.0 libglib2.0-dev
   libgtk-3-dev``.
```
jtojnar: Forgotten closing backticks.

hamster_gtk/helpers.py
```diff
@@ -137,3 +138,62 @@ def get_config_instance(fallback_config_instance, app_name, file_name):
...
+    Returns:
+        re.MatchObject or dict: ``re.MatchObject`` if ``raw=True``, else ``dict``.
+            Returning the ``re.MatchObject`` is particularly useful if one is
+            interested in the groups ``span``s.
+
+    Note:
+        This is not at all about providing valid facts or even raw facts. This function
+        is only trying to extract whatever information can be matched to its various
+        groups (aka 'segments').
+        This not withstanding can however be the bases for future implementations
```
jtojnar: Notwithstanding is a single word, however is probably redundant, and bases should be basis.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+
+from gi.repository import GObject, Gtk
+from orderedset import OrderedSet
+from six import text_type
+from hamster_gtk.helpers import _u
+from hamster_gtk import helpers
+
+
+class RawFactEntry(Gtk.Entry):
+    """A custom entry widgets that provides ``faw fact`` specific autocompletion behaviour."""
```
jtojnar: widgets → widget, faw → raw

hamster_gtk/helpers.py
```diff
@@ -137,3 +138,62 @@ def get_config_instance(fallback_config_instance, app_name, file_name):
...
+    datetime_regex = r'({date}|{time}|{date} {time})'.format(date=date_regex, time=time_regex)
+    # Please note the trailing whitespace!
+    timeinfo_regex = r'({relative} |{datetime} |{datetime} - {datetime} )'.format(
+        datetime=datetime_regex,
+        relative=relative_time_regex
+    )
+    # This is the central place where we define which characters are viable for
+    # our various segments.
+    # Please note that this is also where we define each segments 'separator'.
+    activity_regex = r'[^@:]+'
```
jtojnar: I am still not fully convinced this is the way to go. [Partial matches](https://pypi.python.org/pypi/regex) could be also considered.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+            else:
+                result = string
+            return result
+
+        if not self.match:
+            return
+
+        match = self.match
+        segment = self.current_segment
+        if add_prefix:
```
jtojnar: This will always be `True` as the argument is shadowed by the helper function.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+        self._controller.signal_handler.connect('facts-changed', self._on_facts_changed)
+        self.set_completion(RawFactCompletion(self._controller))
+        # The re.match instance of the current string or None
+        self.match = None
+        # Identifier for the segment the cursor is currently in. None if no
+        # match is available.
+        self.current_segment = None
+        self.connect('changed', self._on_changed)
+
+    def replace_segment_string(self, segment_string, add_prefix=True):
```
jtojnar: What is the purpose of the `add_prefix` when it is never used?

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+        old_string = _u(self.get_text())
+        new_string = '{}{}{}'.format(
+            old_string[:match.start(segment)],
+            segment_string,
+            old_string[match.end(segment):],
+        )
+        self.set_text(new_string)
+        self.set_position(match.start(segment) + len(segment_string))
+
+    def get_segment_text(self, remove_prefix=True):
```
jtojnar: What is the purpose of the `remove_prefix` when it is never used?

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+        category of `` #`` for tags.
+        """
+        def remove_prefix(segment, string):
+            result = string
+            if segment == 'category':
+                result = string[1:]
+            return result
+
+        result = self.match.group(self.current_segment)
+        if remove_prefix:
```
jtojnar: The argument is shadowed by the function.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def get_segment_text(self, remove_prefix=True):
+        """
+        Return the string for the segment given by ``self.current_segment``.
+
+        Please be advised that his text may include a segments 'prefix', eg.g. ``@`` for
+        category of `` #`` for tags.
+        """
+        def remove_prefix(segment, string):
+            result = string
```
jtojnar: In `add_prefix` this is in the else block, it should be unified.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+            return result
+
+        result = self.match.group(self.current_segment)
+        if remove_prefix:
+            result = remove_prefix(self.current_segment, result)
+        return result
+
+
+    # Callbacks
+    def _on_facts_changed(self, evtl):
```
jtojnar: evtl?

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+        """Callback triggered when facts have changed."""
+        self.set_completion(RawFactCompletion(self._controller))
+
+    def _on_changed(self, widget):
+        """
+        Callback triggered whenever entry text is changed.
+
+        Its main task is to keep track of which segment of the raw fact string the
+        user is currently editing. For this the whole string is inspecred and matched
```
jtojnar: inspecred

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+        against our regex. By coparing the current cursor position with individual
```
jtojnar: coparing

hamster_gtk/misc/widgets/raw_fact_entry.py
jtojnar: Also, should not this method by private?

hamster_gtk/misc/widgets/raw_fact_entry.py
jtojnar: Also, should not this method by private?

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+class RawFactCompletion(Gtk.EntryCompletion):
+    """
+    Return a completion instance to match 'activity@category' strings.
+
+    Returns:
+        Gtk.EntryCompletion: Completion instance.
+    """
+
+    def __init__(self, controller, *args, **kwargs):
+        """instantiate class."""
```
jtojnar: Should begin with capital letter.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+        self.segment_models = {
+            'activity': self._activities_model,
+            'category': self._categories_model,
+        }
+
+    def _get_stores(self):
+        # [FIXME]
+        # This needs to be improved. Right now activity and categories can be
+        # duplicates.
+        activities, categories = [], []
```
jtojnar: Why not just use set again?

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -0,0 +1,229 @@
...
+        categories_store = Gtk.ListStore(GObject.TYPE_STRING)
+        for category in OrderedSet(categories):
+            categories_store.append([category])
+        return (activities_store, categories_store)
+
+    def _get_activities(self):
+        """
+        Return all activities that should be considered for autocompletion.
+
+        This is the place were we define which reference frame should be used
```
jtojnar: were → where

hamster_gtk/helpers.py
jtojnar: Additionally, I think `a#b@c` should be parsed as `{'activity': 'a', 'tags': '#b@c'}`, that is `#` should be banned from `activity`.

jtojnar (changes requested):

jtojnar:
Additionally, the separate autocomplete for activities and categories makes no sense to me. After all, each activity has a fixed category. I like how the code is structured, though; it will come in handy for tag autocompletion.

hamster_gtk/helpers.py
elbenfreund: Is there a particular argument to be had for "banning ``#`` within the activity name"?
I personally do not care either way, but I think a general rule "allow as much as possible, unless for technical reasons" makes sense?

jtojnar: Otherwise, `a #b` would be ambiguous, I think.

elbenfreund:
With regards to splitting autocomplete for activity and category:

Please not that it *is true* that each ``Actitivity`` has at max. one category assigned, but ``Activity.name`` ``is not`` a unique (natural) key! Only ``Activity.name + Activity.category`` as a composite key are unique.

Now, this may be something one could do different, but for now this is how ``hamster-lib`` (following legacy hamster for compatibility) handles things.

Now, that does not by itself mean that autocompletion could not behave differently to its current suggestion. For starters, when looking up categories, it may very well make sense to just look up categories that "match" the already known activity (instead of all fetched ones). Or one threats the "activity string" as the whole of "name + category" and looks it up as a whole.
At this point the given solution seemed to strike a balance between flexibility and implementation effort.
I think it would be good not to focus too much on user preferences but get the general machinery in place. Details like this can then be easily adapted to RL user expectations.

hamster_gtk/misc/widgets/raw_fact_entry.py
elbenfreund: The original reason was to keep the "order information" so one could use it to prioritize matches. But if we ever implement a more clever matching solution (may in line with legacy hamster) we can just remove the ``set``. So you are right, as it is now its neither here nor there...

hamster_gtk/misc/widgets/raw_fact_entry.py
elbenfreund: I think it is a rather generic method and thought to make it part of the public API. Which is also why I included the ``add_prefix`` in the signature - to increase user ease. But you are right, I can not really come up with a usecase where one would want to deal with prefixing itself (e.g. one would use ``add_prefix=False``). Damn you featuritis. Going to remove it.

hamster_gtk/misc/widgets/raw_fact_entry.py
elbenfreund: Same as ``replace_segment_string``

jtojnar:
Matching against `{name}@{category}` seems to be what hamster 2 is doing. It also feels more natural.

hamster_gtk/misc/widgets/raw_fact_entry.py
elbenfreund: Yes. I can never make up my mind which to find better.

hamster_gtk/helpers.py
elbenfreund: that should be ``Activity.name='a', Activity.tags={b}``, or not?

jtojnar: I think this regex parses it as `Activity(name='a #b')`.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -29,14 +29,50 @@
...
+            ``activity``, ``category``, ``activity+category``, ``tags`` and ``description``.
+        match: A match object instance.
+
+    Returns:
+        tuple: Tuple (int, int) representing the given segments start and end position.
+    """
+    # We can not simply use ``match.span`` as ``activity+category`` is not
+    # represented as such in the match instance.
+    if segment == 'activity+category':
+        start = match.start('activity')
```
jtojnar: This should probably allow matching start of category when activity is not specified. At least this is how it behaves in hamster-master.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -29,14 +29,50 @@
...
+    """
+    Return start and end positions for a given segment.
+
+    Args:
+        segmemt (text_type): The segment we want to know the boundries for. Valid choices are:
+            ``activity``, ``category``, ``activity+category``, ``tags`` and ``description``.
+        match: A match object instance.
+
+    Returns:
+        tuple: Tuple (int, int) representing the given segments start and end position.
```
jtojnar: segment's

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -29,14 +29,50 @@
...
+
+
 class RawFactEntry(Gtk.Entry):
     """A custom entry widgets that provides ``raw fact`` specific autocompletion behaviour."""
 
-    def __init__(self, controller, *args, **kwargs):
-        """Instantiate class."""
+    def __init__(self, controller, split_activity_autocomplete=False, *args, **kwargs):
+        """
+        Instantiate class.
```
jtojnar: Should be initialize.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -95,8 +132,17 @@ def remove_prefix(segment, string):
                 result = string[1:]
             return result
 
-        result = self.match.group(self.current_segment)
-        result = remove_prefix(self.current_segment, result)
+        if self.current_segment is None:
+            return
+
+        if self.current_segment == 'activity+category':
+            result = self.match.group('activity')
```
jtojnar: To allow matching just on categories, something like `result = (self.match.group('activity') or '') + (self.match.group('category') or '')` should be used.

jtojnar (changes requested):

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -39,12 +39,14 @@ def _get_segment_boundaries(segment, match):
...
 
     Returns:
-        tuple: Tuple (int, int) representing the given segments start and end position.
+        tuple: Tuple (int, int) representing the given segment's start and end position.
     """
     # We can not simply use ``match.span`` as ``activity+category`` is not
     # represented as such in the match instance.
     if segment == 'activity+category':
         start = match.start('activity')
+        if start == -1:
```
jtojnar: This should probably be the same as the second condition.

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -136,13 +138,20 @@ def remove_prefix(segment, string):
...
+                    result += category
+            else:
+                result = None
         else:
-            text = self.match.group(self.current_segment)
-            result = remove_prefix(self.current_segment, text)
+            result = self.match.group(self.current_segment)
+            if result:
+                result = remove_prefix(self.current_segment, result)
+            result
```
jtojnar: Forgotten line

elbenfreund:
Ready for you :)

jtojnar (approved):

Issue #125, opened 2016-10-10
Clean up the HeaderBar


jtojnar:
HeaderBar contains too many buttons that will not be used frequently (About, Preferences). We should group them under a menu, like other modern applications do.

![Nautilus](https://cloud.githubusercontent.com/assets/705123/19246275/c540bd36-8f25-11e6-9a0d-8a464ae7a33d.png)

elbenfreund:
I absolutely agree. The current "design" was just a rough "proof of concept" to get started.

PR #159, opened 2017-01-31
[#125] HeaderBar cleanup

jtojnar:
I thought I had already sent a PR.

This PR moves the About and Preferences buttons into a [application menu](https://lazka.github.io/pgi-docs/Gio-2.0/classes/MenuItem.html#Gio.MenuItem). In GTK, the easiest and cleanest way to add a menu is to pack it into a [resource](https://developer.gnome.org/gio/stable/GResource.html#GResource.description) file and load that into the application. As a consequence, this PR also incorporates the existing css into aforementioned resource file.

screenshot from 2017-01-31 10-28-37

Closes: #125


elbenfreund:
I think it would be good to elaborate a bit on the use of GTK resource

hamster_gtk/hamster_gtk.py
```diff
@@ -198,6 +164,24 @@ def save_config(self, config):
         config_helpers.write_config_file(cp_instance, 'hamster-gtk', 'hamster-gtk.conf')
         self.controller.signal_handler.emit('config-changed')
 
+    def _create_actions(self):
+        """Create actions and add register them in the application."""
```
elbenfreund: remove 'add'

hamster_gtk/resources/gtk/menus.ui
```diff
@@ -0,0 +1,21 @@
+<interface>
```
elbenfreund, line 1: Should we specify XML standard version and encoding?

jtojnar:
Updated. Now, thinking about it, compiling resources should be incorporated into the build/packaging process. Where exactly should I add it? As a dependency to some Makefile rule?

elbenfreund (changes requested):

elbenfreund:
We probably also need to make sure that resources are build as part of the release process /dist target as well as that the resulting resource file is included in the package.

hamster_gtk/resources/gtk/menus.ui
jtojnar: For XML 1.0, the header is optional.  It defaults to some Unicode encoding.

jtojnar:
I still do not understand the build process. Do I modify Makefile, MANIFEST.in or setup.py?

elbenfreund:
All of the above. :)
1. We need to declare ``build-resources`` a dependency for the dist target.
2. We need to [include the resulting resource file](http://python-packaging-user-guide.readthedocs.io/distributing/#package-data) as part of the packaging in ``setup.py``
3. We most likely need to inform our manifest.in about the new file so it doesn't issue warning.

If you prefer I could address those packaging related issues though...
Also, it may be better to not just dump the resources file in the main python code dir but use a dedicated build dir similar to what we do in ``hamster-shell-extension``.

jtojnar:
If you could tackle it, that would be great. I only see the extension creating a zipball:

https://github.com/projecthamster/hamster-shell-extension/blob/master/build.sh#L3

elbenfreund:
done. Some tests seem still trivially broken tough...

setup.py
```diff
@@ -48,6 +48,6 @@
     ''',
 
     package_data={
-        'hamster_gtk': ('resources/hamster-gtk.gresource',),
+        'hamster_gtk': ['resources/hamster-gtk.gresource'],
```
elbenfreund: What the reason for  using a ``list`` instead of a ``tuple``?

jtojnar: sdist is concatenating some list with the `package_data` items so with tuple it fails on the following error:

```
TypeError: can only concatenate list (not "tuple") to list
```

elbenfreund (approved):

PR #160, opened 2017-01-31
Use intersphinx

jtojnar:
Intersphinx adds links to documentation of classes from third party packages. For example when a method returns `match` object, there will be a link to [python docs](https://docs.python.org/3.6/library/re.html#match-objects).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -38,7 +38,7 @@ import hamster_gtk
 # -- General configuration ---------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-#needs_sphinx = '1.0'
+needs_sphinx = '1.5'
 
 # Add any Sphinx extension module names here, as strings. They can be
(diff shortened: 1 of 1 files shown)


elbenfreund:
Cool. Can you create create an issue documenting why this should be done please? Or, at minimum provide some brief context in the PR description.

Also, why the increase of the min sphinx version? Glancing over the docs I could not find an obvious reason.

jtojnar:
If I recall correctly, there were some problems with `napoleon_include_init_with_doc` before 1.5, cannot find it at the moment.

Updated the description.

elbenfreund:
Ok. will merge then. Ideally issues like the one motivating the upgrade to sphinx 1.5 would have their separate ticket and PR.

jtojnar:
Well, we have already upgraded so it should not be a problem.

elbenfreund:
Of cause, it just seems valuable information if we know we *need* a minimal version as opposed to make sure that we can run with the most recent.

Issue #76, opened 2016-09-19
Add support for tagging


riker09:
In the original Hamster app it was possible to categorize items by a simple tagging feature. Will there be something similar for hamster-gtk? I did a quick glimpse at the app, but was not able to find something alike.

I think tags are a great way to keep tracked items organized. Plus if I ever wanted to automatically log my tracked times to Jira, Redmine, etc. a tag with the issue key would be very helpful.

elbenfreund:
Yes. I personally do not use tags but since its last release ``hamster-lib`` actually has support for them. ``hamster-gtk`` currently does not do anything with them though and there is no representation for them at all.
Future versions of ``hamster-gtk`` (probably not the very next one though) will provide support for tags, so keep an eye out for them :)

PR #161, opened 2017-01-31
Show tags in Overview

jtojnar:
Replace pseudo-tag in Overview dialogue with real tags. Adding tags is not actually supported until hamster-lib starts parsing them, though.

### DIFF
--- a/hamster_gtk/overview/widgets/fact_grid.py
+++ b/hamster_gtk/overview/widgets/fact_grid.py
@@ -213,13 +213,7 @@ class FactBox(Gtk.Box):
         return activity_label
 
     def _get_tags_widget(self, fact):
-        """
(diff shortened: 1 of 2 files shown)


elbenfreund (approved):

elbenfreund (dismissed):
This PR breaks ``TestFactBox.test__get_tags_widget`` in ``tests.overview.test_widgets``.

elbenfreund (approved):

elbenfreund:
lgtm. Will merge after rebase

elbenfreund:
This should deal with #76 as well right? Whilst it will not actually be possible to create tags ``hamster-gtk`` has all its cogs in place once ``hamster-lib`` gets to it?!

jtojnar:
Once #150 is merged, we will need it again.

Issue #149, opened 2016-11-15
TestHamsterSpinButton.test_params_constructor_overlapping_bounds fails randomly


elbenfreund:
TestHamsterSpinButton.test_params_constructor_overlapping_bounds somtimes fails to raises the expected ValueError.

```
==================================================================================================== FAILURES =====================================================================================================
________________________________________________________________________ TestHamsterSpinButton.test_params_constructor_overlapping_bounds _________________________________________________________________________

self = <tests.preferences.widgets.test_hamster_spin_button.TestHamsterSpinButton object at 0x7f249aef6ad0>, simple_adjustment = SimpleAdjustment(min=0, max=0, step=98)

    def test_params_constructor_overlapping_bounds(self, simple_adjustment):
        """Passing lower bound greater than upper bound to the constructor should fail."""
        new_adj = SimpleAdjustment(min=simple_adjustment.max, max=simple_adjustment.min,
            step=simple_adjustment.step)
        with pytest.raises(ValueError):
>           HamsterSpinButton(new_adj)
E           Failed: DID NOT RAISE

tests/preferences/widgets/test_hamster_spin_button.py:42: Failed
```

This does happen ever so often (rough guess: 1 out of 6). Whilst it is probably a trivial thing, I didn't spend time investigating so far.....

jtojnar:
Probably happens when the bounds are equal.

PR #162, opened 2017-01-31
[#149] Fix test_params_constructor_overlapping_bounds sometimes failing

jtojnar:
The test swaps upper and lower bounds on the SimpleAdjustment it receives and then expects the HamsterSpinButton to raise. Sometimes, however, the same bounds would be generated and the button would fail to raise.

This commit ensures the generated adjustment never has the same bounds.

Closes: #149

### DIFF
--- a/tests/preferences/widgets/conftest.py
+++ b/tests/preferences/widgets/conftest.py
@@ -72,9 +72,11 @@ def adjustment(request, numbers):
 
 @pytest.fixture
 def simple_adjustment(request, faker):
-    """Return three random numbers in an ascending order."""
+    """Return a ``SimpleAdjustment``."""
     a = faker.random_number()
     b = faker.random_number()
+    if a == b:
+        b += 1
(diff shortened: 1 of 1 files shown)


tests/preferences/widgets/conftest.py
```diff
@@ -75,6 +75,8 @@ def simple_adjustment(request, faker):
     """Return three random numbers in an ascending order."""
```
elbenfreund: Should read  "Return a ``SimpleAdjustment``". Which of cause *is* three random numbers but this seems more consistent with the fixture name and actually returned class.

elbenfreund (changes requested):

tests/preferences/widgets/conftest.py
jtojnar: Oops, I noticed it and then forgot about it.

elbenfreund (approved):

PR #166, opened 2017-02-14
Remove compiled 'gresource' from repository

elbenfreund:
The compile gresource file has been added to the repository by mistake. This is fixed by this PR.

Additionally this PR makes sure that assets needed to compile a gresource file are included in the source distribution package. Because this now means that ``hamster-gtk.gresource`` is not in the dvcs but in the ``manifest.in`` an exception had to be added for ``check-manifest``.


setup.cfg
```diff
@@ -27,9 +27,11 @@ exclude = build/*.py,docs/*.py,*/.ropeproject/*
 ignore = E128
 max-line-length = 99
 
+[check-manifest]
+ignore = hamster_gtk/resources/hamster-gtk.gresource
```
jtojnar, line 31: Why is this needed?

elbenfreund: Because otherwise ``check-manifest`` will notice that ``hamster-gtk.gresource`` is part of the sdist but not of the repository and hence will fail. As this is exactly what we want however, this line is needed.

jtojnar: How will it become part of the sdist?

elbenfreund: because it is included in the manifest

jtojnar (approved):

Issue #168, opened 2017-02-16
Make Activities used to autocomplete against configurable


elbenfreund:
Right now raw fact autocompletion uses only today's activities when fetching ``recent activities``. For obvious reasons it would be nice to have this timeframe used to fetch "recent activities" configuable in the client.

What is needed:
- [ ] A new client config setting
- [ ] Integration of this confifg setting into the preferences dialog
- [ ] ``RawFacCompletion._get_activities`` needs to use this setting

The new setting should be an ``int`` of "days before today" so we will fetch all activities from then till today.

Issue #21, opened 2016-06-17
Provide option to use gconf for config storage


elbenfreund:

jtojnar:
http://zderadicka.eu/gsettings-flexible-configuration-system/

PR #172, opened 2017-02-16
[#168] Make activities to be matched by autocomplete configurable 

elbenfreund:
This PR introduces a new config key ``autocomplete_activities_offset`` which is expected to be an integer.
When autocompleting raw facts a set of 'recent activities' is fetched. This new config value will determine the number of days before today that should be the start for this collection of 'recent activities'.
We also add another new config key ``autocomplete_split_activity`` which is a boolean and governs if a typed string will be matched against ``activity.name`` and ``activity.category.name`` individually or in conjunction.

Closes: #168

### DIFF
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -35,7 +35,7 @@ import hamster_lib
 # under python 2 is practically non existing and manual encoding is not easily
 # possible.
 from backports.configparser import SafeConfigParser
-from gi.repository import Gio, Gdk, GObject, Gtk
+from gi.repository import Gdk, Gio, GObject, Gtk
 from hamster_lib.helpers import config_helpers
 from six import text_type
 
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -104,7 +104,7 @@ class MainWindow(Gtk.ApplicationWindow):
         )
 
         # Set tracking as default screen at startup.
-        self.add(TrackingScreen(self.app.controller, app.config))
+        self.add(TrackingScreen(self.app))
 
 
 # [FIXME]
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -104,7 +104,7 @@ class MainWindow(Gtk.ApplicationWindow):
         )
 
         # Set tracking as default screen at startup.
-        self.add(TrackingScreen(self.app.controller))
+        self.add(TrackingScreen(self.app.controller, app.config))
 
 
 # [FIXME]
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -144,6 +144,7 @@ class HamsterGTK(Gtk.Application):
         # Yes this is redundent, but more transparent. And we can worry about
         # this unwarrented assignment once it actually matters.
         self._config = self._reload_config()
+        self.config = self._config
 
         self._create_actions()
 
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -276,6 +276,8 @@ class HamsterGTK(Gtk.Application):
             'tmpfile_path': os.path.join(appdirs.user_data_dir, 'hamster-gtk.tmp'),
             'db_engine': 'sqlite',
             'db_path': os.path.join(appdirs.user_data_dir, 'hamster-gtk.sqlite'),
+            # Frontend
+            'autocomplete_activities_offset': 30,
         }
 
     def _config_to_configparser(self, config):
@@ -306,6 +308,9 @@ class HamsterGTK(Gtk.Application):
         def get_db_path():
             return text_type(config['db_path'])
 
+        def get_autocomplete_activities_offset():
+            return text_type(config['autocomplete_activities_offset'])
+
         cp_instance = SafeConfigParser()
         cp_instance.add_section('Backend')
         cp_instance.set('Backend', 'store', get_store())
@@ -314,6 +319,10 @@ class HamsterGTK(Gtk.Application):
         cp_instance.set('Backend', 'tmpfile_path', get_tmpfile_path())
         cp_instance.set('Backend', 'db_engine', get_db_engine())
         cp_instance.set('Backend', 'db_path', get_db_path())
+
+        cp_instance.add_section('Frontend')
+        cp_instance.set('Frontend', 'autocomplete_activities_offset',
+                        get_autocomplete_activities_offset())
 
         return cp_instance
 
@@ -364,11 +373,15 @@ class HamsterGTK(Gtk.Application):
                 })
             return result
 
+        def get_autocomplete_activities_offset():
+            return cp_instance.getint('Frontend', 'autocomplete_activities_offset')
+
         result = {
             'store': get_store(),
             'day_start': get_day_start(),
             'fact_min_delta': get_fact_min_delta(),
             'tmpfile_path': get_tmpfile_path(),
+            'autocomplete_activities_offset': get_autocomplete_activities_offset(),
         }
         result.update(get_db_config())
         return result
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -278,6 +278,7 @@ class HamsterGTK(Gtk.Application):
             'db_path': os.path.join(appdirs.user_data_dir, 'hamster-gtk.sqlite'),
             # Frontend
             'autocomplete_activities_offset': 30,
+            'autocomplete_split_activity': False,
         }
 
     def _config_to_configparser(self, config):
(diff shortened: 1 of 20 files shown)


jtojnar:
How will this work in the context of hamster-lib’s config manager (i.e. [LIB-235](https://projecthamster.atlassian.net/browse/LIB-235))?

jtojnar:
Actually, I think the config helpers are still something that every client should implement by itself. For example hamster-gtk should go for #21.

hamster_gtk/hamster_gtk.py
```diff
@@ -275,6 +276,8 @@ def _get_default_config(self):
             'tmpfile_path': os.path.join(appdirs.user_data_dir, 'hamster-gtk.tmp'),
             'db_engine': 'sqlite',
             'db_path': os.path.join(appdirs.user_data_dir, 'hamster-gtk.sqlite'),
+            # Frontend
+            'autocomplete_activities_offset': 30,
```
jtojnar: I do not really like this name. Something like `activity_autocomplete_range` sounds slightly better but still not optimal.

elbenfreund:
Whilst I agree this is something we need to talk about I would like to avoid getting distracted from the PR at hand.
As far as this PR is concerned ``hamster-gtk`` ist config file based and this PR provides a way to deal with this in a consistent way. If there is to emerge a change / extention in the way configuration is done this will be a different matter.

Do you have any other feedback/comments/improvements about this PR or does this mean you are ok with the actual code?

ps: with regards to LIB-235: as far as this PR is concerned this issue is still unresolved. However, even if ``hamster-gtk`` would use the infrastructure provided by ``hamster-lib`` this particular config setting is a "frontend" setting and as such totally unrelated to whatever ``hamster-lib`` would aim to provide...

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -265,8 +268,9 @@ def _get_activities(self):
         for autocomplete suggestions.
         """
         today = datetime.date.today()
+        start = today - datetime.timedelta(self._config['autocomplete_activities_offset'])
```
jtojnar: It might be better to explicitly specify that it is the days argument.

hamster_gtk/hamster_gtk.py
```diff
@@ -363,11 +373,15 @@ def get_db_config():
                 })
             return result
 
+        def get_autocomplete_activities_offset():
+            return cp_instance.getint('Frontend', 'autocomplete_activities_offset')
```
jtojnar: When existing config is used, selfoss crashes on`backports.configparser.NoSectionError: No section: 'Frontend'`

hamster_gtk/misc/widgets/raw_fact_entry.py
```diff
@@ -60,22 +60,24 @@ def _get_segment_boundaries(segment, match):
 class RawFactEntry(Gtk.Entry):
     """A custom entry widgets that provides ``raw fact`` specific autocompletion behaviour."""
 
-    def __init__(self, controller, split_activity_autocomplete=False, *args, **kwargs):
+    def __init__(self, controller, config, split_activity_autocomplete=False, *args, **kwargs):
```
jtojnar: Changes in the preferences do not propagate – restart is needed.

hamster_gtk/preferences/preferences_dialog.py
```diff
@@ -73,6 +73,10 @@ def __init__(self, parent, app, initial, *args, **kwargs):
                 ('db_path', (_('DB _Path'), ComboFileChooser())),
                 ('tmpfile_path', (_('_Temporary file'), ComboFileChooser())),
             ]))),
+            (_('Misc'), LabelledWidgetsGrid(collections.OrderedDict([
```
jtojnar: I would rather see Miscellaneous.

hamster_gtk/hamster_gtk.py
elbenfreund: I do not like the name either, but fail to come up with somehting better.
I think ``activity_autocomplete_range`` is worthe than starting with "autocomplete" as the latter limits the "namespace" more than "activity" and hence already makes the name more specific towards its use. Also it is not used  to "autocomplete activities" but to "fetch activities for autocompletion in general". I agree though that ``..._range`` or maybe even better ``..._days`` would be clearer.

hamster_gtk/hamster_gtk.py
elbenfreund: Yes. This is somewhat expected / desired. Right now all config keys are mandatory. so a preexisting config without all relevant keys is invalid. As we are not API stable at this point its acceptable that we crash with "invalid" older configs.

Before a major release we probably should provide default fallbacks for non supplied config keys though.

elbenfreund: Yes. This is somewhat expected / desired. Right now all config keys are mandatory. so a preexisting config without all relevant keys is invalid. As we are not API stable at this point its acceptable that we crash with "invalid" older configs.

Before a major release we probably should provide default fallbacks for non supplied config keys though.

elbenfreund:
Ready again

jtojnar:
When trying to save a fact, I get “'HamsterControl' object has no attribute 'controller'”.

elbenfreund:
yes. sorry about that. should be fixed

hamster_gtk/preferences/preferences_dialog.py
```diff
@@ -76,6 +78,8 @@ def __init__(self, parent, app, initial, *args, **kwargs):
             (_('Miscellaneous'), LabelledWidgetsGrid(collections.OrderedDict([
                 ('autocomplete_activities_offset', (_("Autocomplete Activities Offset"),
                     HamsterSpinButton(SimpleAdjustment(0, GObject.G_MAXDOUBLE, 1)))),
+                ('autocomplete_split_activity', (_("Autocomplete Split Activity"),
```
jtojnar: “Autocomplete activities and categories separately” would be probably more meaningful.

hamster_gtk/preferences/preferences_dialog.py
```diff
@@ -32,7 +32,9 @@
 from hamster_gtk.preferences.widgets import (ComboFileChooser,
                                              HamsterComboBoxText,
                                              HamsterSpinButton,
-                                             SimpleAdjustment, TimeEntry)
+                                             HamsterCheckButton,
```
jtojnar: Isort

hamster_gtk/preferences/widgets/__init__.py
```diff
@@ -21,4 +21,5 @@
 from .config_widget import ConfigWidget  # NOQA
 from .hamster_combo_box_text import HamsterComboBoxText  # NOQA
 from .hamster_spin_button import HamsterSpinButton, SimpleAdjustment  # NOQA
+from .hamster_check_button import HamsterCheckButton  # NOQA
```
jtojnar: Isort

hamster_gtk/preferences/widgets/hamster_check_button.py
```diff
@@ -0,0 +1,58 @@
...
+#
+# 'hamster-gtk' is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with 'hamster-gtk'.  If not, see <http://www.gnu.org/licenses/>.
+
+"""This module provides ComboBoxText widget extended to store preferences."""
```
jtojnar: wrong description

elbenfreund:
Ready

hamster_gtk/preferences/preferences_dialog.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, parent, app, initial, *args, **kwargs):
                 ('tmpfile_path', (_('_Temporary file'), ComboFileChooser())),
             ]))),
             (_('Miscellaneous'), LabelledWidgetsGrid(collections.OrderedDict([
-                ('autocomplete_activities_offset', (_("Autocomplete Activities Offset"),
+                ('autocomplete_activities_range', (_("Autocomplete Activities Range"),
```
jtojnar: Something like “Autocomplete activities newer than” would be more descriptive.

hamster_gtk/hamster_gtk.py
jtojnar: Either way, I would use [singular](https://en.wikipedia.org/wiki/English_plurals#Nouns_used_attributively).

hamster_gtk/preferences/preferences_dialog.py
elbenfreund: While I agree that it may be slightly better it is still not really great. At some point I think some additional (optional/hover...) description would be useful for most settings I guess. I suggest we revisit the naming choices then instead. Also, this may have the benefit of ``hamster-gtk`` having a bigger audience that may help coming up with better descriptions.

elbenfreund:
ready

jtojnar (dismissed):

Issue #175, opened 2017-02-25
Overview does not order dates


elbenfreund:
When the overview lists more than one day, this listing does not seem to order it's days.

PR #185, opened 2017-06-27
Refactor and Improve Overview

elbenfreund:
This PR refactors part of the Overview related codebase (which was in big parts taken from the legacy widgets). As such we improve how totals are computed as well as move the ``get_delta_string`` method to a dedicated helper function.
On the surface level we now sow barcharts for activities and dates in the details panel as well.
Also the general look of the charts has been improved. While still do not focus too much on the aesthetics at this point this should make everything not only more appealing but also useable.

Also:
- Dates in the overview listing are not ordered (#175)

Closes: #175

### DIFF
--- a/hamster_gtk/helpers.py
+++ b/hamster_gtk/helpers.py
@@ -181,7 +181,7 @@ def get_delta_string(delta):
     """
     Return a human readable representation of ``datetime.timedelta`` instance.
 
-    In most contexts its not that useful to present the delta in seconds.
+    In most contexts it is not that useful to present the delta in seconds.
     Instead we return the delta either in minutes or ``hours:minutes`` depending on the
     value.
 
--- a/hamster_gtk/helpers.py
+++ b/hamster_gtk/helpers.py
@@ -25,6 +25,7 @@ import datetime
 import re
 
 import six
+from six import text_type
 
 
 def _u(string):
@@ -174,3 +175,35 @@ def decompose_raw_fact_string(text, raw=False):
     if match and not raw:
         result = match.groupdict()
     return result
+
+
+def get_delta_string(delta):
+    """
+    Return a human readable representation of ``datetime.timedelta`` instance.
+
+    In most contexts its not that useful to present the delta in seconds.
+    Instead we return the delta either in minutes or ``hours:minutes`` depending on the
+    value.
+
+    Args:
+        delta (datetime.timedelta): The timedelta instance to render.
+
+    Returns:
+        text_type: The datetime instance rendered as text.
+
+    Note:
+        So far, this does not account for large deltas that span days and more.
+    """
+    seconds = delta.total_seconds()
+    minutes = int(seconds / 60)
+    if minutes < 60:
+        result = '{} min'.format(minutes)
+    else:
+        result = '{hours:02d}:{minutes:02d}'.format(
+            hours=int(seconds / 3600), minutes=int((seconds % 3600) / 60))
+    return text_type(result)
+
+
+def rgb_to_gtk_rgb(r, g, b):
+    """Map '255' based RGB values to 0-1 based floats expected by cairo."""
+    return float(r) / 255, float(g) / 255, float(b) / 255
--- a/hamster_gtk/overview/dialogs/overview_dialog.py
+++ b/hamster_gtk/overview/dialogs/overview_dialog.py
@@ -171,58 +171,46 @@ class OverviewDialog(Gtk.Dialog):
             We handle totals as part of this method in order to limit the
             amount of iterations over ``self._facts``.
         """
-        # [FIXME]
-        # This is probably useful to other clients as well and worth
-        # considering for hamster-lib.
-        def get_total(totals):
-            """Return a dictionary of max deltas per key."""
-            max_total = defaultdict(datetime.timedelta)
-            # [FIXME]
-            # is this really the *max* total, and not rather the total per key?
-            for key, deltas in totals.items():
-                for delta in deltas:
-                    max_total[key] += delta
-            return max_total
-
+        # All facts for a given date
         facts_by_date = defaultdict(list)
-        date_deltas = defaultdict(list)
+        date_deltas = defaultdict(datetime.timedelta)
+        # All facts for a given category
         facts_by_category = defaultdict(list)
-        category_deltas = defaultdict(list)
+        category_deltas = defaultdict(datetime.timedelta)
+        # All facts for a given activity
         facts_by_activity = defaultdict(list)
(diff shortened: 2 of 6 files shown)


hamster_gtk/overview/dialogs/overview_dialog.py
```diff
@@ -138,12 +138,15 @@ def _on_charts_button(self, button):
...
+            self._charts = Gtk.ScrolledWindow()
+            self._charts.set_min_content_height(200)
+            self._charts.set_min_content_height(400)
+            self._charts.add(widgets.Charts(self._totals))
             self.main_box.pack_start(self._charts, False, False, 0)
             self.show_all()
 
     # [FIXME]
-    # To avoid multiple falls to the backend, maybe some rudimentaty chaching
+    # To avoid multiple calls to the backend, maybe some rudimentaty caching
```
jtojnar: rudimentaty → rudimentary

hamster_gtk/helpers.py
```diff
@@ -174,3 +175,35 @@ def decompose_raw_fact_string(text, raw=False):
     if match and not raw:
         result = match.groupdict()
     return result
+
+
+def get_delta_string(delta):
+    """
+    Return a human readable representation of ``datetime.timedelta`` instance.
+
+    In most contexts its not that useful to present the delta in seconds.
```
jtojnar: its → it is

hamster_gtk/overview/dialogs/overview_dialog.py
```diff
@@ -138,12 +138,15 @@ def _on_charts_button(self, button):
             self._charts.destroy()
             self._charts = False
         else:
-            self._charts = widgets.Charts(self._totals)
+            self._charts = Gtk.ScrolledWindow()
+            self._charts.set_min_content_height(200)
```
jtojnar: Why these magic numbers?

hamster_gtk/overview/widgets/charts.py
```diff
@@ -34,65 +38,77 @@ class Charts(Gtk.Grid):
...
-        for category, total in category_totals.items():
-            category_label = Gtk.Label(category)
-            bar_chart = HorizontalBarChart(total.total_seconds(), max_total.total_seconds())
-            total_label = Gtk.Label(self._get_delta_string(total))
-            grid.attach(category_label, 0, row, 1, 1)
-            grid.attach(bar_chart, 1, row, 1, 1)
-            grid.attach(total_label, 2, row, 1, 1)
-            row += 1
-        return grid
+        self.override_background_color(Gtk.StateType.NORMAL, Gdk.RGBA(1, 1, 1, 1))
```
jtojnar: This is quite unreadable in dark mode:

hamster_gtk/overview/dialogs/overview_dialog.py
```diff
@@ -168,58 +171,46 @@ def _group_facts(self):
...
+            by_activity=facts_by_activity,
+            by_category=facts_by_category,
+            by_date=facts_by_date,
         )
-        return grouped_facts, totals
 
-    def _get_highest_totals(self, dictionary, amount):
+        totals = Totals(activity_deltas, category_deltas, date_deltas)
+
+        Result = namedtuple('Result', ('grouped_facts', 'totals'))
```
jtojnar, line 209: Why is this a named tuple when it is only ever used for simple tuple destructing.

hamster_gtk/overview/widgets/charts.py
```diff
@@ -34,65 +38,77 @@ class Charts(Gtk.Grid):
...
-            result = '{hours:02d}:{minutes:02d}'.format(
-                hours=int(seconds / 3600), minutes=int((seconds % 3600) / 60))
-        return result
+            max_total = max(totals.values())
+            # Sorting a dict like this returns a list of tuples.
+            totals = sorted(totals.items(), key=operator.itemgetter(1),
+                            reverse=True)
+
+        grid = Gtk.Grid()
+        grid.set_column_spacing(5)
```
jtojnar: Is it possible to use CSS for this? Maybe some selector like `#Overview grid>*`?

jtojnar (changes requested):

hamster_gtk/overview/dialogs/overview_dialog.py
elbenfreund: Otherwise the scrolled Window will not expand has the content technically can be displayed in a very small window due to scolling.

jtojnar: I meant why it is not something like `window_width` and `window_height / 2`.

hamster_gtk/overview/dialogs/overview_dialog.py
elbenfreund: Luciano Ramalho (autor of "Fluent Python") [convincingly argues](https://talkpython.fm/episodes/show/24/fluent-python) that functions that return tuples should always return named tuples instead for readability/logging/debugging reasons. As there is no cost associated with it (other than the extra line for setting up the tuple) I came to adapt this as a best practice.

hamster_gtk/overview/widgets/charts.py
elbenfreund: I find the GTK+ docs really lacking in this regard, but as far as my research goes this seems not possible. [This](https://gist.github.com/ptomato/0fb634ef4098bb89026f) lists all supported properties (as well as the the [classes API docs](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Grid.html#style-properties)) and lacks anything related to column spacing :(

Note:
Also see [this](https://stackoverflow.com/questions/32162372/which-gtk-elements-support-which-css-properties#32170049) SE question...

Feedback from #gnome-hackers @ freenode

> (15:15:48) ebassi: The CSS subset we allow in GTK changes the content, not the position of widgets; it controls the minimum size, but not the given size of widgets

elbenfreund:
rdy

hamster_gtk/overview/widgets/charts.py
jtojnar: That looks slightly better but still difficult to read. How do other applications deal with this?

screenshot from 2017-06-30 19-35-42

elbenfreund: I think GTK+ has a rather sophisticated theming support that allows for colors specific to darkmode or something similar.

elbenfreund:
I removed the changes to the colouring as they not really important right now and I would like to move on with this. theming needs to be addressed at some later stage anyway.

jtojnar (approved):

Issue #189, opened 2017-06-29
Do not install tests


jtojnar:
When installing the app using `python setup.py install --root="${pkgdir}" --optimize=1`, the tests will also be installed, even when calling `sed "s/find_packages()/find_packages(exclude=['tests'])/" -i setup.py` beforehand.

This may lead to a conflict when installing another app that contains the same file (e.g. `/usr/lib/python3.6/site-packages/tests/helpers/__init__.py`).

Default build log: https://paste.gnome.org/pgfkanom1
Build log with tests excluded and removed from `MANIFEST.in`: https://paste.gnome.org/pv04lfs9l

jtojnar:
Apparently `tests*` needs to be excluded from `find_packages`:

https://packaging.python.org/tutorials/distributing-packages/?highlight=tests#packages

Now, can we make that default?

elbenfreund:
yes. so to be clear: the reason exclude did not work before was becaue of ``test*`` vs ``tests``?.

jtojnar:
Yes

jtojnar:
Yes `tests` → `tests*`

PR #191, opened 2017-06-30
Exclude tests from installation

jtojnar:
Previously, the tests module was included when installing the application. This could lead to a conflict when another python package also exposed its tests and used a same file, for example `tests/helpers/__init__.py`.

This patch excludes the tests module from installation. The tests module will still be included in the distribution tarballs so that hamster-gtk can be tested before installation.

Closes: #189


elbenfreund (approved):
lgtm

Issue #124, opened 2016-10-09
Use of backports namespace


jtojnar:
For compatibility with Python 2, we are using [configparser](https://pypi.python.org/pypi/configparser) package. It is also used, however, for Python 3, where it is not necessary.

> To use the `configparser` backport instead of the built-in version on both Python 2 and Python 3, simply import it explicitly as a backport:
> 
> ```
> from backports import configparser
> ```
> 
> If you’d like to use the backport on Python 2 and the built-in version on Python 3, use that invocation instead:
> 
> ```
> import configparser
> ```

Why not use the second import? It would make packaging easier on systems that do not provide `python3-configparser` package. This is how the packaging looks like now: [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python-hamster-lib).

elbenfreund:
At the time, using backports seemed the cleaner and more straight forward solution as it makes sure the same package is used with python 2 and 3. i do not have any general objections to using the standard configparser module under python3 if there are no side effects.
If we decide to go down that route, a corresponding issue should be created with `hamster-lib`.

elbenfreund:
An issue with [``hamster-lib``](https://projecthamster.atlassian.net/browse/LIB-236) has been created.

elbenfreund:
Closing as this has been solved in ``hamster-lib==0.13.0``

jtojnar:
This seems to be still true in `hamster-gtk`, try grepping for backports.

PR #192, opened 2017-06-30
Remove the use of backports namespace

jtojnar:
Continuing with LIB-236, we also remove the namespace from hamster-gtk.

Closing: #124


elbenfreund (approved):
Thanks for dealing with this.

Issue #199, opened 2017-07-04
Provide a basic *.desktop file


elbenfreund:
In order for applications to integrate with GNOME and its app menu ``*.desktop`` files are to be used.
While ultimately ``hamster-gtk`` should be packaged by interested distros to fit within their ecosystem we should provide a simple desktop file that allows it users to install the package from the repository and integrate into their DE.

Issue #197, opened 2017-07-03
'make install' needs to make sure resources are build in advance


elbenfreund:
When installing from the repository, we need to make sure that an up-to-date- version of the resource file is present.

Todo: Make the ``install`` target depend on the ``resources`` target.

PR #202, opened 2017-07-04
[#199] Provide Desktop-File

elbenfreund:
This PR provides a simple desktop file for easy GNOME integration

Closes: #197 
Closes: #199

PR #203, opened 2017-07-06
[#199] Provide basic GNOME integration

elbenfreund:
This commit integrates ``hamster-gtk`` as a integrated GNOME
application. In order to do so, a dedicated ``application__id`` is
assigned (which allows ``Gtk.Application`` to manage instance
uniqueness). Furthermore we now provide a ``*.desktop`` file with basic
metadata that is used to 'register' the app with GNOME.
To make this process as easy as possible, we also provide a new make
tarket ``register-gtk`` which uses the ``desktop-file-install`` helper
tool to handle copying the desktop file to the appropiate location.

On top of those changes this commit also makes sure that ``make
install`` will build a ``gresources`` file beforehand.

Closes: #197
Closes: #199

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,24 +19,25 @@ BROWSER := python -c "$$BROWSER_PYSCRIPT"
 
 help:
 	@echo "Please use 'make <target>' where <target> is one of"
-	@echo "   clean         to remove all build, test, coverage and Python artifacts"
-	@echo "   clean-build   to remove build artifacts"
-	@echo "   clean-pyc     to remove Python file artifacts"
+	@echo "   clean          to remove all build, test, coverage and Python artifacts"
+	@echo "   clean-build    to remove build artifacts"
+	@echo "   clean-pyc      to remove Python file artifacts"
 	@echo "   clean-docs"
-	@echo "   clean-test    to remove test and coverage artifacts"
-	@echo "   lint          to check style with flake8"
-	@echo "   test          to run tests quickly with the default Python"
-	@echo "   test-all      to run tests on every Python version with tox"
-	@echo "   coverage      to check code coverage quickly with the default Python"
+	@echo "   clean-test     to remove test and coverage artifacts"
+	@echo "   lint           to check style with flake8"
+	@echo "   test           to run tests quickly with the default Python"
+	@echo "   test-all       to run tests on every Python version with tox"
+	@echo "   coverage       to check code coverage quickly with the default Python"
 	@echo "   coverage-html"
 	@echo "   codecov"
-	@echo "   develop       to install (or update) all packages required for development"
-	@echo "   docs          to generate Sphinx HTML documentation, including API docs"
-	@echo "   isort         to run isort on the whole project."
-	@echo "   resources     to generate GTK resources"
-	@echo "   release       to package and upload a release"
-	@echo "   dist          to package"
-	@echo "   install       to install the package to the active Python's site-packages"
+	@echo "   develop        to install (or update) all packages required for development"
+	@echo "   docs           to generate Sphinx HTML documentation, including API docs"
+	@echo "   isort          to run isort on the whole project."
+	@echo "   resources      to generate GTK resources"
+	@echo "   release        to package and upload a release"
+	@echo "   dist           to package"
+	@echo "   install        to install the package to the active Python's site-packages"
+	@echo "   register-gnome to register the package as a GNOME Application."
 
 clean: clean-build clean-pyc clean-test
 
@@ -114,5 +115,8 @@ dist: resources
 	python setup.py bdist_wheel
 	ls -l dist
 
-install: clean
+install: clean resources
 	python setup.py install
+
+register-gnome:
+	desktop-file-install misc/org.projecthamster.hamster-gtk.desktop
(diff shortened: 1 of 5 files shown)


elbenfreund:
I created this new PR instead of the messed up #202 one. Hope this makes review easier :)

Makefile
```diff
@@ -114,5 +115,8 @@ dist: resources
 	python setup.py bdist_wheel
 	ls -l dist
 
-install: clean
+install: clean resources
 	python setup.py install
+
+register-GNOME:
```
jtojnar: The make targets are usually lowercase. We do not capitalize HTML either.

misc/org.projecthamster.hamster-gtk.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Name=hamster-gtk
```
jtojnar, line 2: Maybe “Hamster Time Tracker” would be more user friendly.

misc/org.projecthamster.hamster-gtk.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
```
jtojnar, line 1: From what have I seen, the directory containing the desktop file is usually called `data`:

* [Nautilus](https://github.com/GNOME/nautilus/tree/master/data)
* [Weather](https://github.com/GNOME/gnome-weather/tree/master/data)
* [Builder](https://github.com/GNOME/gnome-builder/tree/master/data)

misc/org.projecthamster.hamster-gtk.desktop
```diff
@@ -0,0 +1,9 @@
...
+Name=hamster-gtk
+Exec=hamster-gtk
+Type=Application
+Terminal=false
+Encoding=UTF-8
+Categories=GTK;GNOME;Utility;
+Version=1.0
+Name[de]=hamster-gtk
```
jtojnar: The localized attributes are usually next to the original attribute.

Also, they are often added during build from gettext files. We should probably leave it for i18n anyway.

elbenfreund: I will remove them for now then. I am not sure about placing them next to the original though. Ive seen both. And in case of several dozen translations having this list before the remaining "metadata" may be inconvenient for manual inspection.

misc/org.projecthamster.hamster-gtk.desktop
elbenfreund: For now I would like to be as specific as possible. And whilst the name certainly is not set in stone, it should not be too generic there may be several "hamster time tracker" frontends in future and we should give the impression that the GTK one is somehow special.

misc/org.projecthamster.hamster-gtk.desktop
elbenfreund: yes, but I do not like this pattern too much. The way I see it ``data`` should be for application specific data that is actually used by the app. What the current ``misc`` dir contains however is just helpers for the deployment environment that have nothing to do with the app in a stricter sense. These are just things that should make integrating the app into a wider environment easier. I really don't mind a different name than ``misc`` though, its not that important to me at this stage - just ``data`` is probably not really my cup of tea.

jtojnar: The application specific data are usually called resources. Though, I do not really care much, just pointing out the conventions I inferred.

misc/org.projecthamster.hamster-gtk.desktop
jtojnar: IMHO, the user should not care about the frontend library, their software repository should offer them the most fitting variant. The details might be listed in the About dialogue.

elbenfreund: IMHO "frontend client"  is all the user cares about. Instead of "the image opener" we call distinct packages different names even if they do the same and may even use the same libs for functionality. As there are already ppl working on a QT version it is easy to see that one repo may offer multiple packages at the same time that are "hamster time trackers" in their own right. The name should reflect that they are but also how they differ I think. As I said: I am happy to discuss this once it is more relevant, right now I think it would be nice to wrap things up :)

jtojnar: I prefer non-generic names for software but if an organization creates an app with multiple different frontends, it makes sense for all of them to share the name. If the QT frontend is developed by a third party not connected to projecthamster.org then using a different name might be okay.

elbenfreund: I would like to keep it like this, at least for now.

elbenfreund:
ready for review again :)

jtojnar:
For reference: [Install a desktop file](https://developer.gnome.org/platform-overview/unstable/dev-launching-desktop.html.en)

README.rst
```diff
@@ -25,6 +25,9 @@ You have been warned.
 Dependencies
 -------------
 
+If you want to use the ``make register-GNOME`` target ``desktop-file-install`` is
```
jtojnar: The make target should be lowercase.

jtojnar (changes requested):
Just a final touch, also the manifest file should be updated

jtojnar (approved):

PR #205, opened 2017-07-06
[#136] Recent activities in 'tracking screen'

elbenfreund:
Closes: #136


elbenfreund:
should be branch ``feature/136`` not ``139``. Will open a new one with the correct branch.

PR #208, opened 2017-07-08
[#136] Add 'recent activities' to tracking screen

elbenfreund:
This adds a 'recent activities' area to the tracking screen that can be used to take previous activities as templates for new raw fact strings or start tracking them right away. The entire functionality is optional (on by default right now) and can be configures via preferences.

The 'recent activities' box size can be adjusted by specifying the amount of activities it shall be able to show. All activities that not fit are available via scrolling.

Closes: #136

### DIFF
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -116,6 +116,10 @@ class SignalHandler(GObject.GObject):
     Once signals have been 'registered' here you can ``emit`` or ``connect`` to
     them via its class instances.
     """
+    # [TODO]
+    # Explain semantics of each signal
+    # [TODO]
+    # Add signals for all changed hamster-lib objects?
 
     __gsignals__ = {
         str('facts-changed'): (GObject.SIGNAL_RUN_LAST, None, ()),
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -433,12 +433,9 @@ class HamsterGTK(Gtk.Application):
 
     def _get_config_from_file(self):
         """
-        Return a config dictionary from acp_instanceg file.
+        Return a config dictionary from app_instance file.
 
-        If there is none create a default config file. This methods main job is
-        to convert strings from the loaded ConfigParser File to appropiate
-        instances suitable for our config dictionary. The actual data retrival
-        is provided by a hamster-lib helper function.
+        If there is none create a default config file.
 
         Returns:
             dict: Dictionary of config key/values.
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -116,6 +116,7 @@ class SignalHandler(GObject.GObject):
     Once signals have been 'registered' here you can ``emit`` or ``connect`` to
     them via its class instances.
     """
+
     # [TODO]
     # Explain semantics of each signal
     # [TODO]
@@ -259,6 +260,7 @@ class HamsterGTK(Gtk.Application):
         """Reload configuration from designated store."""
         config = self._get_config_from_file()
         self._config = config
+        self.config = config
         return config
 
     def _config_changed(self, sender):
@@ -283,6 +285,8 @@ class HamsterGTK(Gtk.Application):
             # Frontend
             'autocomplete_activities_range': 30,
             'autocomplete_split_activity': False,
+            'tracking_show_recent_activities': True,
+            'tracking_recent_activities_items': 6,
         }
 
     def _config_to_configparser(self, config):
@@ -319,6 +323,12 @@ class HamsterGTK(Gtk.Application):
         def get_autocomplete_split_activity():
             return text_type(config['autocomplete_split_activity'])
 
+        def get_tracking_show_recent_activities():
+            return text_type(config['tracking_show_recent_activities'])
+
+        def get_tracking_recent_activities_items():
+            return text_type(config['tracking_recent_activities_items'])
+
         cp_instance = SafeConfigParser()
         cp_instance.add_section('Backend')
         cp_instance.set('Backend', 'store', get_store())
@@ -333,6 +343,10 @@ class HamsterGTK(Gtk.Application):
                         get_autocomplete_activities_range())
         cp_instance.set('Frontend', 'autocomplete_split_activity',
                         get_autocomplete_split_activity())
+        cp_instance.set('Frontend', 'tracking_show_recent_activities',
+                        get_tracking_show_recent_activities())
+        cp_instance.set('Frontend', 'tracking_recent_activities_items',
+                        get_tracking_recent_activities_items())
 
         return cp_instance
 
@@ -389,6 +403,12 @@ class HamsterGTK(Gtk.Application):
         def get_autocomplete_split_activity():
             return cp_instance.getboolean('Frontend', 'autocomplete_split_activity')
 
+        def get_tracking_show_recent_activities():
+            return cp_instance.getboolean('Frontend', 'tracking_show_recent_activities')
+
+        def get_tracking_recent_activities_items():
+            return int(cp_instance.get('Frontend', 'tracking_recent_activities_items'))
+
         result = {
             'store': get_store(),
             'day_start': get_day_start(),
@@ -396,6 +416,8 @@ class HamsterGTK(Gtk.Application):
             'tmpfile_path': get_tmpfile_path(),
             'autocomplete_activities_range': get_autocomplete_activities_range(),
             'autocomplete_split_activity': get_autocomplete_split_activity(),
+            'tracking_show_recent_activities': get_tracking_recent_activities_items(),
+            'tracking_recent_activities_items': get_tracking_recent_activities_items(),
         }
         result.update(get_db_config())
         return result
--- a/hamster_gtk/hamster_gtk.py
+++ b/hamster_gtk/hamster_gtk.py
@@ -286,7 +286,7 @@ class HamsterGTK(Gtk.Application):
             'autocomplete_activities_range': 30,
             'autocomplete_split_activity': False,
             'tracking_show_recent_activities': True,
-            'tracking_recent_activities_items': 6,
+            'tracking_recent_activities_count': 6,
         }
 
     def _config_to_configparser(self, config):
@@ -326,8 +326,8 @@ class HamsterGTK(Gtk.Application):
         def get_tracking_show_recent_activities():
             return text_type(config['tracking_show_recent_activities'])
 
-        def get_tracking_recent_activities_items():
-            return text_type(config['tracking_recent_activities_items'])
+        def get_tracking_recent_activities_count():
+            return text_type(config['tracking_recent_activities_count'])
 
         cp_instance = SafeConfigParser()
         cp_instance.add_section('Backend')
@@ -345,8 +345,8 @@ class HamsterGTK(Gtk.Application):
                         get_autocomplete_split_activity())
         cp_instance.set('Frontend', 'tracking_show_recent_activities',
                         get_tracking_show_recent_activities())
-        cp_instance.set('Frontend', 'tracking_recent_activities_items',
-                        get_tracking_recent_activities_items())
+        cp_instance.set('Frontend', 'tracking_recent_activities_count',
+                        get_tracking_recent_activities_count())
 
         return cp_instance
 
@@ -406,8 +406,8 @@ class HamsterGTK(Gtk.Application):
         def get_tracking_show_recent_activities():
             return cp_instance.getboolean('Frontend', 'tracking_show_recent_activities')
 
-        def get_tracking_recent_activities_items():
-            return int(cp_instance.get('Frontend', 'tracking_recent_activities_items'))
+        def get_tracking_recent_activities_count():
+            return int(cp_instance.get('Frontend', 'tracking_recent_activities_count'))
 
         result = {
             'store': get_store(),
@@ -416,8 +416,8 @@ class HamsterGTK(Gtk.Application):
             'tmpfile_path': get_tmpfile_path(),
             'autocomplete_activities_range': get_autocomplete_activities_range(),
             'autocomplete_split_activity': get_autocomplete_split_activity(),
-            'tracking_show_recent_activities': get_tracking_recent_activities_items(),
-            'tracking_recent_activities_items': get_tracking_recent_activities_items(),
+            'tracking_show_recent_activities': get_tracking_show_recent_activities(),
+            'tracking_recent_activities_count': get_tracking_recent_activities_count(),
         }
         result.update(get_db_config())
         return result
--- a/hamster_gtk/helpers.py
+++ b/hamster_gtk/helpers.py
@@ -24,6 +24,7 @@ from __future__ import absolute_import, unicode_literals
 import datetime
 import re
 
+from orderedset import OrderedSet
 import six
 from six import text_type
 
@@ -177,6 +178,14 @@ def decompose_raw_fact_string(text, raw=False):
     if match and not raw:
         result = match.groupdict()
     return result
+
+
+# [TODO]
+# Oncec LIB-251 has been fixed this should no longer be needed.
+def get_recent_activities(controller, start, end):
+    """Return a list of all activities logged in facts within the given timeframe."""
+    recent_activities = [fact.activity for fact in controller.facts.get_all(start=start, end=end)]
+    return OrderedSet(recent_activities)
 
 
 def get_delta_string(delta):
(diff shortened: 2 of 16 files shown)


hamster_gtk/tracking/screens.py
```diff
@@ -238,3 +268,119 @@ def _on_start_tracking_button(self, button):
     def _on_raw_fact_entry_activate(self, evt):
         """Callback for when ``enter`` is pressed within the entry."""
         self._start_ongoing_fact()
+
+    def _on_config_changed(self, sender):
+        """Callback triggered when 'config-changed' event fired."""
+        if self._app.config['tracking_show_recent_activities']:
+            # We re-create it even if one existed before necause its parameters
```
jtojnar: necause

hamster_gtk/preferences/preferences_dialog.py
```diff
@@ -80,6 +80,11 @@ def __init__(self, parent, app, initial, *args, **kwargs):
                 ('autocomplete_split_activity',
                  (_("Autocomplete activities and categories separately"),
                   HamsterSwitch())),
+                ('tracking_show_recent_activities',
+                 (_("Allow tracking based on recent activities."),
```
jtojnar: This does not sound clear to me. I would choose something like “Show recent activities for quickly starting tracking”

hamster_gtk/helpers.py
```diff
@@ -177,6 +180,23 @@ def decompose_raw_fact_string(text, raw=False):
...
 
+# [TODO]
+# Oncec LIB-251 has been fixed this should no longer be needed.
+def get_recent_activities(controller, start, end):
+    """Return a list of all activities logged in facts within the given timeframe."""
+    recent_activities = [fact.activity for fact in controller.facts.get_all(start=start, end=end)]
+    return OrderedSet(recent_activities)
+
+
+def serialize_activity(activity):
```
jtojnar: Do not we have something like this somewhere already?

hamster_gtk/hamster_gtk.py
```diff
@@ -279,6 +285,8 @@ def _get_default_config(self):
             # Frontend
             'autocomplete_activities_range': 30,
             'autocomplete_split_activity': False,
+            'tracking_show_recent_activities': True,
+            'tracking_recent_activities_items': 6,
```
jtojnar: I would use `count` or `number`. `items` is less clear.

hamster_gtk/tracking/screens.py
```diff
@@ -230,6 +239,27 @@ def reset(self):
...
+
+    def _get_recent_activities_widget(self):
+        scrolled_window = Gtk.ScrolledWindow()
+        scrolled_window.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC)
+        grid = RecentActivitiesGrid(self, self._app.controller)
+        # We need to 'show' the grid early in order to make sure space is
+        # allocated to its children so they actually have a height that we can
+        # use.
+        grid.show_all()
+        # We fetch an arbitrary Button as height-reference
```
jtojnar: this is really ugly and crashes when there are no items:

```
$ env XDG_CONFIG_HOME=$HOME/.dev_config XDG_DATA_HOME=$HOME/.dev_local hamster-gtk
Hamster-GTK started.
Traceback (most recent call last):
  File "/nix/store/fzrxvdjas3icqb2n5akahvix758jnfvj-python3.6-hamster-gtk/lib/python3.6/site-packages/hamster_gtk/hamster_gtk.py", line 213, in _activate
    self.window = MainWindow(app)
  File "/nix/store/fzrxvdjas3icqb2n5akahvix758jnfvj-python3.6-hamster-gtk/lib/python3.6/site-packages/hamster_gtk/hamster_gtk.py", line 107, in __init__
    self.add(TrackingScreen(self.app))
  File "/nix/store/fzrxvdjas3icqb2n5akahvix758jnfvj-python3.6-hamster-gtk/lib/python3.6/site-packages/hamster_gtk/tracking/screens.py", line 49, in __init__
    self.start_tracking_view = StartTrackingBox(self._app)
  File "/nix/store/fzrxvdjas3icqb2n5akahvix758jnfvj-python3.6-hamster-gtk/lib/python3.6/site-packages/hamster_gtk/tracking/screens.py", line 195, in __init__
    self.recent_activities_widget = self._get_recent_activities_widget()
  File "/nix/store/fzrxvdjas3icqb2n5akahvix758jnfvj-python3.6-hamster-gtk/lib/python3.6/site-packages/hamster_gtk/tracking/screens.py", line 255, in _get_recent_activities_widget
    child = grid.get_children()[1]
IndexError: list index out of range
Traceback (most recent call last):
  File "/nix/store/fzrxvdjas3icqb2n5akahvix758jnfvj-python3.6-hamster-gtk/lib/python3.6/site-packages/hamster_gtk/hamster_gtk.py", line 218, in _activate
    app.add_window(self.window)
TypeError: Argument 1 does not allow None as a value
Hamster-GTK shut down.
```

hamster_gtk/tracking/screens.py
```diff
@@ -238,3 +268,119 @@ def _on_start_tracking_button(self, button):
...
+
+    def refresh(self, sender=None):
+        """Clear the current content and re-populate and re-draw the widget."""
+        helpers.clear_children(self)
+        self._populate()
+        self.show_all()
+
+    def _populate(self):
+        """Fill the widget with rows per activity."""
+        def add_row_widgets(row_counter, activity):
```
jtojnar: `row_counter` → `row_index`/`row_number`; counter implies an object that counts

jtojnar (changes requested):

hamster_gtk/helpers.py
```diff
@@ -177,6 +180,23 @@ def decompose_raw_fact_string(text, raw=False):
     return result
 
 
+# [TODO]
+# Oncec LIB-251 has been fixed this should no longer be needed.
+def get_recent_activities(controller, start, end):
+    """Return a list of all activities logged in facts within the given timeframe."""
+    recent_activities = [fact.activity for fact in controller.facts.get_all(start=start, end=end)]
+    return OrderedSet(recent_activities)
```
jtojnar: The activities are returned in a convoluted order. It should be reversed prior to removing duplicates.

hamster_gtk/tracking/screens.py
```diff
@@ -238,3 +268,119 @@ def _on_start_tracking_button(self, button):
...
+                button.connect('clicked', self._on_start_button, activity)
+                return button
+
+            self.attach(get_label(activity), 0, row_counter, 1, 1)
+            self.attach(get_copy_button(activity), 1, row_counter, 1, 1)
+            self.attach(get_start_button(activity), 2, row_counter, 1, 1)
+
+        today = datetime.date.today()
+        start = today - datetime.timedelta(90)
+        activities = reversed(helpers.get_recent_activities(self._controller, start, today))
```
jtojnar: Do we really want all the activities and scroll instead of displaying the `tracking_recent_activities_items`?

jtojnar (changes requested):

hamster_gtk/tracking/screens.py
```diff
@@ -45,7 +45,8 @@ def __init__(self, app, *args, **kwargs):
         self.set_transition_type(Gtk.StackTransitionType.SLIDE_UP)
         self.set_transition_duration(1000)
         self.current_fact_view = CurrentFactBox(self._app.controller)
-        self.current_fact_view.connect('tracking-stopped', self.update)
+        self.current_fact_view.con275G
```
jtojnar: wah

hamster_gtk/helpers.py
elbenfreund: I think earlier ``hamster-lib`` branches had something like this. Also the overview has its own "serialize activity" code which is different however:
- it deals with legacy hamerest representation of ``activity.category=None``. This is something that I would be willing to throw out of the window.
- It uses `` - `` as seperator between ``activity.name`` and ``category``. I think this is better looking for the overview. What do you think?

So this helper provides a generic serialization utilizing the standart ``@`` seperator. As far as I can tell current versions of ``hamster-lib`` or ``hamster-gtk`` do not have such a function yet. But share the feeling that we had this around for a while in various incarnations...

jtojnar: * Wait, do we intend to not support category-less facts?
* I actually prefer the at sign for consistency or habit.

hamster_gtk/helpers.py
elbenfreund: I am not sure I do understand. Could you elaborate please?

jtojnar: It should be `OrderedSet(reversed(recent_activities))`, otherwise you will get the activities in the order they were first used, not the last.

```
facts     1 2 1 3 2 1 3  output   reversed
current   x x   x       → 1 2 3  → 3 2 1
expected          x x x → 3 1 2  → 2 1 3
```

hamster_gtk/helpers.py
elbenfreund: - No we do and want to support category less facts. I am not quite sure that the way legacy hamster did is really where we want to go.. 
- Does that mean you would prefer ``@`` in the overview as well? At this stage I do not care about this detail very much as we will have to revisit the general layout at some later stage anyway...

jtojnar: Yes, having `@` in overview is what I am accustomed to.

hamster_gtk/tracking/screens.py
elbenfreund: I would prefer such a more general solution. There is nothing preventing us from limiting the results to just ``tracking_recent_activities_items`` (via config) if we want, but especially given that we already have a solution I do not see the harm. 

We **need** scrolling anyway (because we can not make any guarantee that ``tracking_recent_activities_items`` will fit in the space available. In light of that, it seems we get the extra activities for free if the user bothers scolling.

What would be the downside?

jtojnar: Accidental scrolling to the other end and performance.

hamster_gtk/helpers.py
elbenfreund: Thank you for the clarification. To be honest, right now ``FactManager.get_all()`` does not commit to any ordering at all. I created an [issue with ``hamster-lib``](https://projecthamster.atlassian.net/browse/LIB-259) to improve this.
For now I will implement a workaround simple sorting call as part of the prelimery ``get_recent_activities`` helper which will hopefully address your concerns.

hamster_gtk/tracking/screens.py
elbenfreund: - "accidental scrolling": Im not sure I see the use case. especially as there is the posibility of a scrollbox even if if we just fetch the number of activities defined in the config.
- by default we just fetch todays facts. this should net us no more than a couple of dozens facts at most I figure. I doubt there is any concievable performance impact on the DB or the GUI rendering. Unless I'm missing something this seems like a case of premature optimization.

hamster_gtk/helpers.py
elbenfreund: Alrighty. Will adjust this then :) Im cool either way tbh.

hamster_gtk/tracking/screens.py
jtojnar: * You scroll somewhere accidentally and since the list is long you do not have a simple way of returning back – you cannot just quickly scroll down and when you find out that you are on the wrong side quickly scroll to the correct one.
* I see more than just today’s facts. In fact there does not seem to be any limit.

hamster_gtk/tracking/screens.py
elbenfreund: I agree that the way we derive our min_height hint is far from elegant (but fixed value layouts are far worse even). I am absolutely open to better approaches! Empty children are definitely a nice catch and have been addressed in the latest version.

hamster_gtk/tracking/screens.py
elbenfreund: You are [right](https://github.com/projecthamster/hamster-gtk/blob/feature/136/hamster_gtk/tracking/screens.py#L354)!

The idea was to make the timedelta adjustable via a config setting, but i forgot to actually do that :(
If done that way it is up to the user to figure what timeframe is viable to them. We probably should point out the performance consideration though... 

It would probably be best to leave this additional customization to a seperate issue and use ``start=today`` for now...

tests/test_hamster-gtk.py
```diff
@@ -14,9 +14,19 @@
 class TestHamsterGTK(object):
     """Unittests for the main app class."""
 
-    def test_instantiation(self):
-        """Make sure class instatiation works as intended."""
-        app = hamster_gtk.HamsterGTK()
+    def test_instantiation(self, config):
+        """
+        Make sure class instatiation works as intended.
```
jtojnar: instatiation → instantiation

elbenfreund:
@jtojnar I addressed most of your concerns and CI is passing again (finally). Do you think you will find some time to review it again?

The only big issue left open is [this](https://github.com/projecthamster/hamster-gtk/pull/208#discussion_r139186084). I am absolutely open to improve the current hackish solution if you have better idea.

hamster_gtk/helpers.py
```diff
@@ -196,10 +197,34 @@ def get_recent_activities(controller, start, end):
...
+def serialize_activity(activity, separator='@', none_category='not categorized'):
+    """
+    Provide a serialized string version of an activity.
+
+    Args:
+        activity (Activity): ``Activity`` instance to serialize.
+        separator (str, optional): ``string`` used to separate ``activity.name`` and
+            ``category.name``. The separator will be omitted if ``none_categoty=''`` and
+            ``activity.category=None``. Defaults to ``@``.
+        none_category (str, optional): ``string`` to represent the 'lack of a cactegory' for an
```
jtojnar: `cactegory`

hamster_gtk/helpers.py
```diff
@@ -177,6 +182,54 @@ def decompose_raw_fact_string(text, raw=False):
     return result
 
 
+# [TODO]
+# Once LIB-251 has been fixed this should no longer be needed.
+def get_recent_activities(controller, start, end):
+    """Return a list of all activities logged in facts within the given timeframe."""
+    # [FIXME]
+    # This manual sorting within python is of cause less than optimal. We stick
```
jtojnar: of cause → of course

hamster_gtk/helpers.py
```diff
@@ -177,6 +182,54 @@ def decompose_raw_fact_string(text, raw=False):
...
+
+
+def serialize_activity(activity, separator='@', none_category='not categorized'):
+    """
+    Provide a serialized string version of an activity.
+
+    Args:
+        activity (Activity): ``Activity`` instance to serialize.
+        separator (str, optional): ``string`` used to separate ``activity.name`` and
+            ``category.name``. The separator will be omitted if ``none_categoty=''`` and
```
jtojnar: none_categoty → none_category

tests/test_hamster-gtk.py
```diff
@@ -14,9 +14,19 @@
...
+
+        We actually test against a monkeypatched class in order to avoid the
+        config loading machinery as this would access the user data on fs.
+        The relevant skiped methods need to be tested separately.
```
jtojnar: skipped

hamster_gtk/tracking/screens.py
jtojnar: Maybe add a todo comment and go with this.

jtojnar (approved):
Several typos, otherwise looks good from a shallow review (ignoring my bloat concerns)

hamster_gtk/helpers.py
```diff
@@ -177,6 +182,54 @@ def decompose_raw_fact_string(text, raw=False):
...
+    # with it for now as this is just a preliminary workaround helper anyway and
+    # effective sorting will need to be implemented by the storage backend in
+    # ``hamster-lib``.
+    facts = sorted(controller.facts.get_all(start=start, end=end),
+        key=operator.attrgetter('start'), reverse=True)
+    recent_activities = [fact.activity for fact in facts]
+    return OrderedSet(recent_activities)
+
+
+def serialize_activity(activity, separator='@', none_category='not categorized'):
```
jtojnar: We probably do not want to serialize uncategorized activities with the “not categorized” category.

screenshot from 2018-05-04 14-40-33

If you copy/start them, the fake category will become real.

elbenfreund: good catch. :+1:

hamster_gtk/tracking/screens.py
elbenfreund: I created #224  as I would like to avoid todo comments.

hamster_gtk/tracking/screens.py
```diff
@@ -238,3 +270,119 @@ def _on_start_tracking_button(self, button):
...
+            """
+            Add a set of widgets to a specific row based on the activity passed.
+
+            Args:
+                row_counter (int): Which row to add to.
+                activity (hamster_lib.Activity): The activity that is represented by this row.
+            """
+            def get_label(activity):
+                """Label representing the activity/category combination."""
+                label = Gtk.Label(helpers.serialize_activity(activity, none_category=''))
```
jtojnar: You will still need to remove `none_category` from here.

Issue #212, opened 2017-08-15
Fix how 'TestEditFactDialog.test_updated_fact_same' handles seconds


elbenfreund:
``hamster-lib`` has improved the way it [handles seconds](https://projecthamster.atlassian.net/browse/LIB-254). As a consequence, this test fails now (running ``hamster-lib >=0.13.2``) as it includes special hacks to account for the previous limitations.
The test should be adjusted to remove the workaround.

PR #213, opened 2017-08-15
[#212] Fix test to play nice with 'hamster-lib >= 0.13.2'

elbenfreund:
This PR removes the workaround from the test that was used to make it work with ``hamster-lib <= 0.13.2``.

Closes: #212

### DIFF
--- a/tests/misc/test_dialogs.py
+++ b/tests/misc/test_dialogs.py
@@ -13,7 +13,6 @@ import datetime
 
 from freezegun import freeze_time
 from gi.repository import Gtk
-import pytest
 
 from hamster_gtk import helpers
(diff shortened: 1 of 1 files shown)


tests/misc/test_dialogs.py
```diff
@@ -184,16 +184,10 @@ def test_updated_fact_same(self, dummy_window, fact):
         Make sure the property returns Fact with matching field values.
 
         We need to jump through some extra hoops because we the current
```
jtojnar: What hoops? “the current implementation will always set the edited fact to today” Will it?

elbenfreund:
rdy :)

jtojnar (approved):

Issue #214, opened 2017-08-15
Running tests with recent requirements fails


elbenfreund:
```
py27 runtests: commands[1] | make coverage

coverage run -m pytest tests

Traceback (most recent call last):

  File "/home/travis/build/projecthamster/hamster-lib/.tox/py27/lib/python2.7/site-packages/_pytest/config.py", line 367, in _importconftest

    mod = conftestpath.pyimport()

  File "/home/travis/build/projecthamster/hamster-lib/.tox/py27/lib/python2.7/site-packages/py/_path/local.py", line 662, in pyimport

    __import__(modname)

  File "/home/travis/build/projecthamster/hamster-lib/.tox/py27/lib/python2.7/site-packages/_pytest/assertion/rewrite.py", line 216, in load_module

    py.builtin.exec_(co, mod.__dict__)

  File "/home/travis/build/projecthamster/hamster-lib/.tox/py27/lib/python2.7/site-packages/py/_builtin.py", line 221, in exec_

    exec2(obj, globals, locals)

  File "<string>", line 7, in exec2

  File "/home/travis/build/projecthamster/hamster-lib/tests/conftest.py", line 15, in <module>

    register(lib_factories.CategoryFactory)

  File "/home/travis/build/projecthamster/hamster-lib/.tox/py27/lib/python2.7/site-packages/pytest_factoryboy/fixture.py", line 60, in register

    deps = get_deps(factory_class, model_name=model_name)

  File "/home/travis/build/projecthamster/hamster-lib/.tox/py27/lib/python2.7/site-packages/pytest_factoryboy/fixture.py", line 154, in get_deps

    for attr, value in factory_class.declarations(factory_class._meta.postgen_declarations).items()

AttributeError: 'FactoryOptions' object has no attribute 'postgen_decla
```

The underlying problem is a recent change in ``factory-boy``s codebase.

PR #215, opened 2017-08-15
[#214] Specicfy 'factory-boy' version in requirements

elbenfreund:
Our explicit test requirement ``pytest-factoryboy`` pulls the latest
``factory-boy`` version by default. This latest version however has bus
that break out test setup which is why this commit explicitly specifies
the last working version.

Closes: #214

### DIFF
--- a/requirements/test.pip
+++ b/requirements/test.pip
@@ -11,6 +11,9 @@
 check-manifest==0.35
 coverage==4.3.4
 doc8==0.7.0
+# This is needed to prevent ``pytest-factoryboy`` to pull a more recent (faulty)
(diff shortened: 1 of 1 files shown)


requirements/test.pip
```diff
@@ -11,6 +11,7 @@
 check-manifest==0.35
 coverage==4.3.4
 doc8==0.7.0
+factory-boy==2.8.1
```
jtojnar: Could you add comment mentioning the bug here?

elbenfreund:
rdy

jtojnar (approved):

